PR #120, opened 2024-07-10
docs(circom-types): added documentation, modified public interfaces, removed unused code

romanmarkusholler:


circom-types/src/groth16/zkey.rs
```diff
@@ -90,19 +68,20 @@ struct Section {
...
+/// Represents a zkey in the format defined by circom. Implements [`ZKey::from_reader`] to deserialize a zkey from a reader.
 #[derive(Clone)]
 pub struct ZKey<P: Pairing> {
     pk: ProvingKey<P>,
     matrices: ConstraintMatrices<P::ScalarField>,
 }
 
 #[derive(Debug, CanonicalSerialize, CanonicalDeserialize)]
-pub struct OurMatrix<F: CanonicalSerialize + CanonicalDeserialize> {
+pub(crate) struct OurMatrix<F: CanonicalSerialize + CanonicalDeserialize> {
```
dkales, line 79: all of these are breaking changes and require a ! in the conventional commit message

Super nice would be to split the documentation and visibility changes into 2 commits, and have the visibility change be:
`refactor!: remove internal structs from public crate interface`

dkales: for this one i will just squash it with a single refactor! commit that also does documentation as a side-effect

dkales (approved):

circom-types/src/groth16/zkey.rs
romanmarkusholler: Got it, thank you!

PR #122, opened 2024-07-12
feat!: Add possibility to use different curves, Shamir for proof generation and translating Rep3 witnesses to Shamir.

rw0x0:


dkales (approved):
LGTM

PR #124, opened 2024-07-15
Parse of hex strings for split-input + Get specific output from SharedWitness

### LINKED ISSUE #92: Feature Request: SplitInput cannot parse hexadecimal field elements
The SplitInput command takes an input.json and secret shares the input. At the moment, it is not possible to provide hexadecimal values.

For that to work, the parse_field function in co-circom.rs must be rewritten.

### LINKED ISSUE #93: Feature Request: Easy way to get specific output from witness
At the moment the all outputs and public inputs are a vec of field elements. It would be nice to have a method to retrieve a specific public input/output

fnieddu:
resolves #92 
resolves #93


dkales (approved):
LGTM

PR #128, opened 2024-08-01
feat: add deserialization of plonk circom types

0xThemis:
Adds functionality to deserialize the following PLONK artifacts form circom:

- proof
- verification key
- zkey


circom-types/src/groth16/r1cs.rs
dkales, line 1: I think from an organization standpoint r1cs should not be below groth16, since it is not only for that, just leave it at the top lvl

circom-types/src/plonk/mod.rs
dkales, line 1: Would like it better if we just re-export the types here with `pub use` instead of nesting deeper for often only a single type, and make the mods private

circom-types/src/plonk/verification_key.rs
```diff
@@ -0,0 +1,239 @@
+//! This module defines the [`JsonVerificationKey`] struct that implements de/serialization using [`serde`] and the [`From`] trait for Plonk.
+
+use ark_ec::pairing::Pairing;
+
+use serde::{Deserialize, Serialize};
+
+use crate::traits::{CircomArkworksPairingBridge, CircomArkworksPrimeFieldBridge};
+
+/// Represents a verification key in JSON format that was created by circom. Supports de/serialization using [`serde`].
```
dkales, line 9: Does the verification key not need a separate format like the proof? Or other way round: why does the proof need 2 different types and this not?

circom-types/src/plonk/zkey.rs
```diff
@@ -0,0 +1,526 @@
...
+        let mut lagrange = Vec::with_capacity(n_public);
+        for _ in 0..n_public {
+            lagrange.push(Self::evaluations(domain_size, &mut reader)?);
+        }
+        Ok(lagrange)
+    }
+
+    fn taus<R: Read>(domain_size: usize, mut reader: R) -> ZKeyParserResult<Vec<P::G1Affine>> {
+        let mut p_tau = vec![];
+        //TODO: why domain size + 6?
```
dkales, line 208: do we have an answer here?

circom-types/src/plonk/zkey.rs
```diff
@@ -0,0 +1,526 @@
...
+            s.spawn(|_| p_tau = Some(Self::taus(domain_size, &mut t_section)));
+        });
+        Ok(Self {
+            n_vars,
+            n_public,
+            domain_size,
+            n_additions,
+            n_constraints,
+            verifying_key: header.verifying_key,
+            additions: additions.unwrap()?,
```
dkales, line 286: maybe add a comment why these unwraps are ok

dkales (changes requested):
This also is a breaking change since it modifies the public interface of various crates.

circom-types/src/groth16/r1cs.rs
0xThemis: Moved it back - we may need a structural overhaul in the near future I think

circom-types/src/plonk/mod.rs
0xThemis: Did that, but only for this file. Maybe do that also for the other ones in a future PR?

circom-types/src/plonk/zkey.rs
0xThemis: Unfortunately, not.

circom-types/src/plonk/zkey.rs
0xThemis: Done

dkales (approved):

PR #133, opened 2024-08-06
fix: updated deny.toml advisories and licenses

fabian1409:


deny.toml
```diff
@@ -57,6 +59,8 @@ allow = [
     # Required for ring
     "OpenSSL",
     "CC0-1.0",
+    "GPL-3.0",
+    "MPL-2.0",
 ]
```
dkales, line 64: we only want to allow these for the very specific crates that need them, not in general.

I think there is an exception part for licenses: https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html#the-exceptions-field-optional

dkales (changes requested):

dkales (approved):

PR #145, opened 2024-08-09
feat: faster reading of circom types when dangerous feature

### LINKED ISSUE #139: FeatureRequest: Add config to not check that points are on curve when deser circom types
Our trait implementations check that every point we read are on the curve and in the cyclic subgroup. Add a configuration that let's us skip these checks. Also maybe only enable it when "dangerous" is enabled?

0xThemis:
When dangerous feature is enabled, the parser will not check wether points are on curve or in the correct subgroup

closes #139


dkales (changes requested):
I don't think this is the way to do this. A feature should not just remove checks IMO, since any upstream crate can add this feature and maybe invalidate some assumptions of other crates also using this.

I would still expose 2 methods, with one being the `unchecked` one, and only expose the unchecked one when the feature is enabled.

0xThemis:
We can't decide on how to solve this at the moment. As it is not high on our checklist we close it and maybe re-open it later

PR #147, opened 2024-08-09
feat!: Add Plonk to the co-circom binary

rw0x0:


collaborative-circom/src/lib.rs
```diff
@@ -27,6 +27,15 @@ use serde::Deserialize;
 /// A module for file utility functions.
 pub mod file_utils;
 
+/// An enum representing the ZK proof system to use.
+#[derive(Clone, ValueEnum)]
```
dkales, line 31: Is the derive here as expected? IDK why we did not use if for the MPCPrococol and MPCCurve as well?

dkales (changes requested):

collaborative-circom/src/lib.rs
rw0x0: I use the derive macro now where applicable

0xThemis (approved):

dkales (approved):

PR #150, opened 2024-08-12
Refactor/remove plain groth16 prover

0xThemis:
This PR has some clean up.
Most notably it adds a verifier feature for co-Groth16. This means the ark::groth16 dep is only used when having this feature enabled.

Additionally, some refactoring to mirror PLONK. We now only have one single prove call to groth16 (we removed the matrices call). We now only take the ZKey and the witness and read the matrices from there, so we never recompute them.

QUESTION: I do not know if we need the R1CS file. We only use it for the witness extension to read the public inputs. We can also do that with the zkey, but snarkjs needs the r1cs file for witness extension therefore I left it. But we could just remove it....


circom-types/src/groth16/verification_key.rs
```diff
@@ -198,6 +194,7 @@ where
         PreparedVerifyingKey::<P>::from(self)
     }
 }
+    */
```
dkales, line 197: why are we keeping this comment

circom-types/src/groth16/verification_key.rs
```diff
@@ -291,53 +288,53 @@ mod test {
             fs::read_to_string("../test_vectors/Groth16/bls12_381/verification_key.json").unwrap();
         let vk = serde_json::from_str::<JsonVerificationKey<Bls12_381>>(&vk_string).unwrap();
         let alpha_1 = test_utils::to_g1_bls12_381!(
-            "1006274644424409217186953213662503172434575368717179668374437354164299156533023899663544377042948523039106817870257",
-            "1056550161284843386841851874013894300183177686460359778532448293436751746312447247913808276245741832724930728476526"
+            "573513743870798705896078935465463988747193691665514373553428213826028808426481266659437596949247877550493216010640",
```
dkales, line 291: why do we need new KAT vectors?

circom-types/src/groth16/verification_key.rs
```diff
@@ -30,32 +26,32 @@ where
...
-    delta_2: P::G2Affine,
+    pub delta_2: P::G2Affine,
     /// The pairing of α and β of the verification key ∈ Gt
     #[serde(rename = "vk_alphabeta_12")]
     #[serde(serialize_with = "P::serialize_gt::<_>")]
     #[serde(deserialize_with = "P::deserialize_gt_element::<_>")]
-    alpha_beta_gt: P::TargetField,
-    /// Referred to as `gamma_abc_g1` in [`ark_groth16::data_structures::VerifyingKey`] and is used to bind the public inputs to the proof
+    pub alpha_beta_gt: P::TargetField,
+    /// Bind the public inputs to the proof
```
dkales, line 50: Used to bind ...

circom-types/src/groth16/zkey.rs
```diff
@@ -195,98 +215,74 @@ where
...
-
 impl<P: Pairing + CircomArkworksPairingBridge> HeaderGroth<P>
 where
     P::BaseField: CircomArkworksPrimeFieldBridge,
     P::ScalarField: CircomArkworksPrimeFieldBridge,
 {
     fn read<R: Read>(mut reader: &mut R) -> ZKeyParserResult<Self> {
-        // TODO: Impl From<u32> in Arkworks
-        let n8q: u32 = u32::deserialize_uncompressed(&mut reader)?;
+        let _n8q: u32 = u32::deserialize_uncompressed(&mut reader)?;
```
dkales, line 251: Should we assert/err on the expected value of these?

collaborative-plonk/src/round1.rs
```diff
@@ -295,12 +295,13 @@ pub mod tests {
     use std::{fs::File, io::BufReader};
 
     use ark_bn254::Bn254;
-    use circom_types::{groth16::witness::Witness, plonk::ZKey, r1cs::R1CS};
-    use collaborative_groth16::{circuit::Circuit, groth16::SharedWitness};
+    use circom_types::plonk::ZKey;
+    use collaborative_groth16::groth16::SharedWitness;
```
dkales, line 299: did we not want to put the sharedwitness into a separate crate? Other PR, right?

collaborative-plonk/src/round3.rs
```diff
@@ -517,7 +517,8 @@ pub mod tests {
     use std::{fs::File, io::BufReader};
 
     use ark_bn254::Bn254;
-    use circom_types::{groth16::witness::Witness, plonk::ZKey};
+    use circom_types::groth16::Witness;
```
dkales, line 520: This witness type is also no longer `groth16` only, it  should live 1 lvl higher

circom-types/src/groth16/verification_key.rs
0xThemis: removed the comment.

circom-types/src/groth16/verification_key.rs
0xThemis: done

circom-types/src/groth16/zkey.rs
0xThemis: I thought about that but I if everything else is correct it doesnt matter what those values are. Therefore I just ignored them. Maybe put a warning log in there?
We maybe want logging anyways?

collaborative-plonk/src/round1.rs
0xThemis: Yes we did wanted to do that, but the plan was to do that in another PR. Also I didn't want to decide alone how we structure everything

collaborative-plonk/src/round3.rs
0xThemis: moved it

circom-types/src/groth16/verification_key.rs
0xThemis: One test we had did not use the .json verification key to verify, but the verification key in the zkey.
I wanted to mirror the actual workflow, therefore tried to read the verification key, but the verification key in the test folder was no longer the one needed for the verification. Therefore, I extracted it from the zkey, overriding the other test case.

I could just keep both, if prefered

dkales: Nah was just a question, is fine as is

circom-types/src/groth16/zkey.rs
dkales: We should probably add a lot more tracing in general in a new pr

dkales: They still have some expected value though and if they are not that then there is some fuckery going on and we should probably error

dkales (approved):

Issue #143, opened 2024-08-09
Update book with new config structure


dkales:
#132 forgot to update the book docs for the config changes.

PR #156, opened 2024-08-13
refactor: each command has own config controlled via file, env, cli

fabian1409:


fabian1409:
resolves (#143)

collaborative-circom/src/lib.rs
```diff
@@ -97,31 +97,395 @@ impl std::fmt::Display for MPCProtocol {
...
+    };
+}
+
+impl_try_from!(SplitInputCli, SplitInputConfig);
+impl_try_from!(SplitWitnessCli, SplitWitnessConfig);
+impl_try_from!(MergeInputSharesCli, MergeInputSharesConfig);
+impl_try_from!(GenerateWitnessCli, GenerateWitnessConfig);
+impl_try_from!(TranslateWitnessCli, TranslateWitnessConfig);
+impl_try_from!(GenerateProofCli, GenerateProofConfig);
+impl_try_from!(VerifyCli, VerifyConfig);
```
dkales, line 487: I don't think `try_from` should have these side-effects? (loading stuff from env/file, etc)
Maybe just make this a method on the target config struct, also via macro

dkales (changes requested):

dkales (approved):

PR #173, opened 2024-08-19
co-noir types and restructure of repo

0xThemis:


co-noir/co-acvm/src/solver/memory_solver.rs
```diff
@@ -0,0 +1,110 @@
...
+                    "value for mem op must be a degree one univariate polynomial with coef 1 and constant 0"
+                ))
+            }
+        } else {
+            Err(eyre::eyre!(
+                "value for mem op must be a degree one univariate polynomial"
+            ))
+        }?;
+        let read_write = op.operation.q_c.into_repr();
+        //TODO CHECK PREDICATE - do we need to cmux here?
```
dkales, line 73: this hopefully should be always public?

co-noir/co-acvm/src/solver/memory_solver.rs
```diff
@@ -0,0 +1,110 @@
...
+    T: NoirWitnessExtensionProtocol<F>,
+    F: PrimeField,
+{
+    pub(super) fn solve_memory_init_block(
+        &mut self,
+        block_id: BlockId,
+        init: &[Witness],
+    ) -> CoAcvmResult<()> {
+        tracing::trace!("solving memory init block {}", block_id.0);
+        // TODO: should we trust on the compiler here?
```
dkales, line 22: I would say that we can leave this there as inits should not dominate the runtime anyways

co-noir/co-acvm/src/solver.rs
```diff
@@ -0,0 +1,226 @@
...
+                _ => todo!("non assert zero opcode detected, not supported yet"),
+                //Opcode::Call {
+                //    id,
+                //    inputs,
+                //    outputs,
+                //    predicate,
+                //} => todo!(),
+            }
+        }
+        // TODO this is most likely not correct. We just reverse the order of the Vec. Maybe this
```
dkales, line 212: Why is this not fine?

mpc-core/src/protocols/plain.rs
```diff
@@ -637,51 +637,78 @@ impl<F: PrimeField> NoirWitnessExtensionProtocol<F> for PlainDriver<F> {
...
-                F::from(u64::try_from(idx).expect("usize fits into u64")),
-                value,
-            );
-        }
-        lut
+    // we could check if a Vec<F> impl may be faster. Depends on the size of the LUT..
+    type SecretSharedSet = HashSet<F>;
+
+    type SecretSharedMap = HashMap<F, F>;
+
```
dkales, line 672: I think we would need `Vec`s here for correctness anyway, since just because the share is equal does not mean the underling value is equal.

co-noir/co-acvm/src/solver/memory_solver.rs
0xThemis: I am not sure how exactly the predicate gets ejected. Nevertheless, we need to solve the equation with the witness map, therefore it is of type AcvmType, which can either be public or shared.

The write operation is tough because we need to write back the value of the LUT, therefore we need to read from the LUT. Or we extend our write with another condition perform an and and cmux over that.

I think for the moment I will just assert that it must be public...

mpc-core/src/protocols/plain.rs
0xThemis: Also for the plain value driver? The "shares" are F. For the rep3 one we have Vecs.

dkales (approved):

Issue #112, opened 2024-07-10
FeatureRequest: Delegate the additional flags from circom to compiler


0xThemis:
https://docs.circom.io/getting-started/compilation-options/#flags-and-options-related-to-the-r1cs-optimization

PR #177, opened 2024-08-28
feat: additional compiler flags

fabian1409:


fabian1409:
fixes (#112)

book/src/config.md
```diff
@@ -71,6 +71,12 @@ This section configures the co-circom MPC compiler.
 * `version`: Allows leaking of secret values in loops (default: "2.0.0")
 * `link_library`: A list of strings that represent paths for circom to look for library files (`-l` flag in circom) (default: `[]`)
 * `allow_leaky_loops`: Allows leaking of secret values in loops (default: `false`) (*currently not implemented*).
+* `simplification`: Controls the simplification level applied during constraint generation (default: `O3(usize::MAX)`)
```
dkales, line 74: this should be O2?

circom-mpc-compiler/src/lib.rs
```diff
@@ -42,16 +42,44 @@ use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, marker::PhantomData, path::PathBuf};
 
+/// The simplification level applied during constraint generation
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, Eq, PartialEq, PartialOrd, Ord, Hash)]
+pub enum Simplification {
```
dkales, line 47: Probably `SimplificationLevel` is a better name

circom-mpc-compiler/src/lib.rs
```diff
@@ -142,15 +173,19 @@ where
...
+                r
+            } else {
+                0
+            },
             flag_json_sub: false,
             json_substitutions: String::new(),
-            flag_s: false,
-            flag_f: false,
+            flag_s: self.config.simplification == Simplification::O1,
+            flag_f: self.config.simplification == Simplification::O0,
```
dkales, line 184: Are these really exclusive? What happens if both are `true`?

dkales (changes requested):

circom-mpc-compiler/src/lib.rs
fabian1409: They are exclusive in circom (https://github.com/iden3/circom/blob/master/circom/src/input_user.rs#L100C1-L102C77)
If both are passed as args in circom, the "lower" (in this case 0) one will be be used, for co-cricom only one can be set in the config

dkales (approved):

PR #190, opened 2024-09-19
Update README.md

adust09:
fix not working url


dkales (approved):
thanks, probably was some autocomplete...

PR #192, opened 2024-09-20
fix path in shell scripts

adust09:


dkales:
the current shell scripts are indented to be run from the examples directory ATM, but we will update this in the future to be more friendly to run.

PR #208, opened 2024-10-04
feat: Add merging inputs to co-noir binary

rw0x0:


co-noir/co-noir/src/bin/co-noir.rs
```diff
@@ -253,6 +262,62 @@ fn run_split_input(config: SplitInputConfig) -> color_eyre::Result<ExitCode> {
...
+        })
+        .collect::<Result<Vec<_>, _>>()?;
+    let mut result = WitnessMap::new();
+    let mut offset = 0;
+
+    for input_share in input_shares.into_iter() {
+        let mut els = 0;
+        for (wit, share) in input_share.into_iter() {
+            let wit = wit.0 + offset;
+            result.insert(wit.into(), share);
```
dkales, line 304: we should probably emit at least a warning if both parties provide the same input

dkales: also what happens now if the input is not complete, does the witness extension fail nicely then?

dkales (changes requested):

co-noir/co-noir/src/bin/co-noir.rs
rw0x0: there is at this point not really a way to check anymore

dkales (approved):

PR #222, opened 2024-10-14
fix: fixed read task breaking too early, caused error during proof gen

fabian1409:


0xThemis:
Used in #220 - no need to merge

PR #223, opened 2024-10-15
chore: added wait statements to example shell scripts

fabian1409:
sometimes example scripts fail because one party finishes faster than the others and the next commands are executed, leading to address already used errors.

by adding wait statements we make sure that all parties are done before we advance.


0xThemis:
Used in #220 - no need to merge

PR #227, opened 2024-10-17
refactor: replace quinn with tokio_rustls

fabian1409:


mpc-net/src/lib.rs
```diff
@@ -124,185 +116,208 @@ impl MpcNetworkHandler {
...
+    // TODO add stats tracking
     /// Returns the number of sent and received bytes.
-    pub fn get_send_receive(&self, i: usize) -> std::io::Result<(u64, u64)> {
-        let conn = self
-            .connections
-            .get(&i)
-            .ok_or_else(|| io::Error::new(io::ErrorKind::NotFound, "no such connection"))?;
-        let stats = conn.stats();
-        Ok((stats.udp_tx.bytes, stats.udp_rx.bytes))
+    pub fn get_send_receive(&self, _i: usize) -> std::io::Result<(u64, u64)> {
```
dkales, line 223: We probably want this for this version

mpc-core/src/protocols/rep3/network.rs
```diff
@@ -245,47 +247,27 @@ impl Rep3MpcNet {
         let runtime = tokio::runtime::Builder::new_multi_thread()
```
dkales, line 247: do we want to limit the # of worker threads here to 2 or so?

mpc-net/src/lib.rs
```diff
@@ -3,64 +3,79 @@
...
+use tokio_util::codec::{Decoder, Encoder, FramedRead, FramedWrite, LengthDelimitedCodec};
 
 pub mod channel;
 pub mod codecs;
 pub mod config;
 
-/// A warapper for a runtime and a network handler for MPC protocols.
-/// Ensures a gracefull shutdown on drop
+// TODO get this from network config
+const STREAMS_PER_CONN: usize = 8;
```
dkales, line 42: This should be in the config, yeah.

mpc-net/src/lib.rs
```diff
@@ -114,195 +124,336 @@ impl MpcNetworkHandler {
...
+        if let Poll::Ready(Ok(bytes_written)) = &res {
+            self.bytes_written
+                .fetch_add(*bytes_written, Ordering::SeqCst);
+        }
+
+        res
+    }
+
+    fn is_write_vectored(&self) -> bool {
+        self.inner.is_write_vectored()
```
dkales, line 457: this could probably live in a separate module

PR #229, opened 2024-10-17
feat: add feature to support config compressed/uncompressed serialize…

### LINKED ISSUE #187: Make it configurable to send compressed/uncompressed over network
It should be possible to configure whether we want to send field elements in compressed/uncompressed mode. This happens in Shamir/Rep3Network. At the moment we send uncompressed.

As soon as we can switch, we want some benchmarks to decide

SuccinctPaul:
Resolved #187

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ Cargo.lock
 *.pdb
 
 .vscode
+.idea
+.DS_Store
+
 co-noir/co-acvm/src/bin/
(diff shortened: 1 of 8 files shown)


0xThemis:
Hi! Thanks for participating in the project :)
Unfortunately, we already had a similar solution already, but we decided against configuring that behavior with a rust feature flag, as this may break upstream crates.

We'd like to have two methods exposed and let the caller decide which to use.

PR #232, opened 2024-10-22
feat: provide binary to inline circom sources

dkales:
only rudimentary parsing at the moment


dkales:
@romanmarkusholler I think this might already be usable enough, even though it is not very robust in parsing the sources, let me know if something does not work

PR #250, opened 2024-10-30
build(deps): update circom dependencies

### LINKED ISSUE #243: Resolver needs to be "1"
When including the co-Circom library crate, we get a compile error from a dependency introduced by circom. We can only resolve that with setting resolver to "1", which is very annoying when trying to use our lib.

Generally, our dependency management is very annoying if we include it from another crate as there are some dependencies that infect the upstream. We should re-export the dependencies or rewrite the interfaces

dkales:
resolves #243 
- **docs: documentation for internal tuple type**
- **build(deps): update circom compiler version**
- **fix: pass field into circom parser**
- **fix: map new circom type to our own type**
- **fix: get the size from new SizeOption type**
- **fix: handle new AccessType type**


Cargo.toml
```diff
@@ -15,7 +15,7 @@ members = [
     "mpc-net",
     "tests",
 ]
-resolver = "1" # need resolver 1 because of circom upstream
+resolver = "2" # need resolver 1 because of circom upstream
```
0xThemis, line 18: remove the comment

0xThemis (approved):

0xThemis (approved):

PR #255, opened 2024-11-06
fix: update README with correct repo url

toolabi:
This branch should:

- [x] change the old repo url that return "not found" with the new one.


dkales (approved):
ty!

Issue #106, opened 2024-07-09
Feature Request: Cannot merge array inputs


0xThemis:
At the moment, we can only merge inputs from multiple parties if it is a dedicated named input. If a circuit has inputs like:

`signal input in[2]`;

it is not possible to provide the array from different parties

- [ ] Implementation
- [ ] Docs example

dkales:
Use "?" as placeholder for other parties.

fabian1409:
> @fabian1409 is this done?

will implement this next week

PR #260, opened 2024-11-11
feat: add support for merging input arrays

fabian1409:
Adds support for merging input arrays.
The "?" character can be used in arrays to indicate, that this element is provided by another party and will will be replaced during input merging.

fixes (#106)


co-circom/co-circom/src/lib.rs
```diff
@@ -672,11 +676,29 @@ where
...
             shares[2].public_inputs.insert(name.clone(), parsed_vals);
         } else {
-            let [share0, share1, share2] =
-                SerializeableSharedRep3Input::share_rep3(&parsed_vals, &mut rng, seeded, additive);
-            shares[0].shared_inputs.insert(name.clone(), share0);
-            shares[1].shared_inputs.insert(name.clone(), share1);
-            shares[2].shared_inputs.insert(name.clone(), share2);
+            // if all elements are Some, then we can share normally
+            // else we can only share as Vec<Option<T>> and we have to merge unknown inputs later
+            if let Some(parsed_vals) = parsed_vals.iter().copied().collect::<Option<Vec<_>>>() {
```
0xThemis, line 681: use
parser_vals.iter().all(|x|x.is_some()) 
instead of this. Then you don't have to clone everything

0xThemis: you maybe need to flatten the iter also when I think about it when using all

0xThemis (changes requested):

co-circom/co-circom/src/lib.rs
fabian1409: > you maybe need to flatten the iter also when I think about it when using all

is already just a Vec<Option<T>>, so no need to

dkales (approved):

PR #266, opened 2024-11-18
Feat/new accels

dkales:


co-circom/circom-mpc-vm/src/mpc.rs
```diff
@@ -117,4 +117,14 @@ pub trait VmCircomWitnessExtension<F: PrimeField> {
 
     /// num2bits accelerator
     fn num2bits(&mut self, a: Self::VmType, bits: usize) -> Result<Vec<Self::VmType>>;
+
+    /// addbits accelerator
+    ///
+    /// Returns the binary addition result of two bit vectors a and b. The bit vectors are expected to have the same length. Also returns the carry output bit.
+    /// a and b are ordered such that the MSB is at the first index, as well as the result
```
rw0x0, line 124: I think everywhere else we have the first index to be the LSB. What is the design choice here for this?

co-circom/circom-mpc-vm/src/mpc/rep3.rs
```diff
@@ -571,6 +572,38 @@ impl<F: PrimeField, N: Rep3Network> VmCircomWitnessExtension<F>
...
+        let bitlen = a.len();
+        let a = a.into_iter().map(|x| match x {
+            Rep3VmType::Public(x) => promote_to_trivial_share(self.io_context0.id, x),
+            Rep3VmType::Arithmetic(x) => x,
+        });
+        let b = b.into_iter().map(|x| match x {
+            Rep3VmType::Public(x) => promote_to_trivial_share(self.io_context0.id, x),
+            Rep3VmType::Arithmetic(x) => x,
+        });
+        let two = F::one() + F::one();
```
rw0x0, line 591: F::from(2u64)?

co-circom/circom-mpc-vm/src/mpc/rep3.rs
```diff
@@ -571,6 +572,38 @@ impl<F: PrimeField, N: Rep3Network> VmCircomWitnessExtension<F>
...
+
+        let a_sum = a.fold(Rep3PrimeFieldShare::zero_share(), |acc, x| acc * two + x);
```
rw0x0, line 593: don't know if acc+acc might be the more efficent one :)

co-circom/circom-mpc-vm/src/mpc/rep3.rs
```diff
@@ -571,6 +572,38 @@ impl<F: PrimeField, N: Rep3Network> VmCircomWitnessExtension<F>
...
+        let b_sum = b.fold(Rep3PrimeFieldShare::zero_share(), |acc, x| acc * two + x);
+
+        let sum = a_sum + b_sum;
```
rw0x0, line 596: are we sure this never overflows the field? otherwise we can not get the carry and the result might be wrong

co-circom/circom-mpc-vm/src/mpc.rs
dkales: That is the convention that Reclaim is using internally, so I did not want to change it.
Another issue is probably that we do not want to have the accelerators in this trait?

PR #280, opened 2024-11-22
feat!: now can specify whether we want curve checks during zkey deser

0xThemis:
BREAKING CHANGE: changed the traits for circom bridge. Also modified the from_reader impl for the two Zkeys


co-circom/circom-types/src/traits.rs
```diff
@@ -508,7 +524,7 @@ where
     phantom_data: PhantomData<P>,
 }
 
-impl<P: Pairing + CircomArkworksPairingBridge> G2Visitor<P>
+impl<P: Pairing + CircomArkworksPairingBridge> TargetGroupVisitor<P>
```
dkales, line 527: what is this for?

co-circom/co-circom/src/bin/co-circom.rs
```diff
@@ -477,7 +477,10 @@ where
 
     let public_input = match proof_system {
         ProofSystem::Groth16 => {
-            let zkey = Arc::new(Groth16ZKey::<P>::from_reader(zkey_file).context("reading zkey")?);
+            let zkey = Arc::new(
+                Groth16ZKey::<P>::from_reader(zkey_file, CheckElement::Yes)
```
dkales, line 481: this should probably be configurable from env

dkales (approved):

co-circom/circom-types/src/traits.rs
0xThemis: just moving stuff around

co-circom/co-circom/src/bin/co-circom.rs
0xThemis: we agreed to move it to later

PR #290, opened 2024-12-10
feat: implement a radix sort in MPC and use it for range checks in co-noir

rw0x0:


mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+    bits: &[Rep3RingShare<T>],
+    io_context: &mut IoContext<N>,
+) -> IoResult<Vec<Rep3RingShare<T>>>
+where
+    Standard: Distribution<T>,
+{
+    let len = rho.len();
+    debug_assert_eq!(len, rho.len());
+
+    let unshuffled = (0..len as PermRing).collect::<Vec<_>>();
```
0xThemis, line 129: Can we only perm 4GB?

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+    let (perm_a, perm_b) = io_context.rngs.rand.random_perm(unshuffled);
+    let perm: Vec<_> = perm_a
+        .into_iter()
+        .zip(perm_b)
+        .map(|(a, b)| Rep3RingShare::new(a, b))
+        .collect();
+
+    let opened = shuffle_reveal::<PermRing, _>(&perm, rho, io_context)?;
+    let bits_shuffled = shuffle(&perm, bits, io_context)?;
```
0xThemis, line 138: These calls can be parallel if we need performance. Fork the network and do join on two threads. We don't need it in this PR but maybe we can write a comment so that we can do it at some point if we start improving the speed.

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+    let (perm_a, perm_b) = io_context.rngs.rand.random_perm(unshuffled);
+    let perm: Vec<_> = perm_a
+        .into_iter()
+        .zip(perm_b)
+        .map(|(a, b)| Rep3RingShare::new(a, b))
+        .collect();
+
+    let opened = shuffle_reveal(&perm, rho, io_context)?;
+    let bits_shuffled = shuffle_field(&perm, bits, io_context)?;
+    let mut result = vec![Rep3PrimeFieldShare::zero_share(); len];
```
0xThemis, line 164: Same as above (forked io_context)

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+                beta_2.push(a.a);
+            }
+            // first shuffle
+            let mut shuffled_1 = Vec::with_capacity(len);
+            for (pi, alpha) in pi.iter().zip(alpha_1) {
+                let pi_1 = pi.b.0 as usize;
+                shuffled_1.push(beta_2[pi_1] + alpha);
+            }
+            io_context.network.send_next_many(&shuffled_1)?;
+            let delta = io_context.network.recv_prev_many()?;
```
0xThemis, line 375: can be one call 
https://github.com/TaceoLabs/co-snarks/blob/c04d2c0d3172aa790f18f4776ba36b248261565e/mpc-core/src/protocols/rep3/network.rs#L164

0xThemis: Generelly if we have `send_next` and `recv_prev` we can do it with this

0xThemis (changes requested):

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+
+fn gen_perm<F: PrimeField, N: Rep3Network>(
+    inputs: &[FieldShare<F>],
+    io_context: &mut IoContext<N>,
+    bitsize: usize,
+) -> IoResult<Vec<Rep3RingShare<PermRing>>> {
+    let mask = (BigUint::one() << bitsize) - BigUint::one();
+
+    // Decompose
+    // TODO: This step could be optimized (I: Pack the a2b's, II: only reconstruct bitsize bits)
```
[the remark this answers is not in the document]
rw0x0: The second part at least I think makes sense if we do this at some point, so I want to keep it here

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
+//! Sort
```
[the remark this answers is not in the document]
rw0x0: The reason is that I use the rep3ring types internally, and I did not want to have a reference from the rep3 field to rep3 rings.

mpc-core/src/protocols/rep3_ring.rs
```diff
@@ -10,6 +10,7 @@ pub mod binary;
 pub mod casts;
 pub mod conversion;
 mod detail;
+pub mod gadgets;
```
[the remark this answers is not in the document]
rw0x0: as discussed personally, this is something we should think about later on

mpc-core/src/protocols/rep3_ring/gadgets/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Gadgets
+//!
+//! This module contains some commonly used gadgets for the Rep3 protocol.
+
+pub mod sort;
```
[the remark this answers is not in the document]
rw0x0: Same as above

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+    inputs: &[FieldShare<F>],
+    io_context: &mut IoContext<N>,
+    bitsize: usize,
+) -> IoResult<Vec<Rep3RingShare<PermRing>>> {
+    let mask = (BigUint::one() << bitsize) - BigUint::one();
+
+    // Decompose
+    // TODO: This step could be optimized (I: Pack the a2b's, II: only reconstruct bitsize bits)
+    let mut bits = Vec::with_capacity(bitsize);
+    for inp in inputs.iter().cloned() {
```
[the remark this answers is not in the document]
rw0x0: Unfortunately after this function (gen_perm), I need the inputs at another point as well (to apply the perm created in gen_perm). So a clone is unfortunately necessary (with the current a2b_selector implementation that takes a full vector)

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+
+fn inject_bit<F: PrimeField, N: Rep3Network>(
+    inputs: &[Rep3BigUintShare<F>],
+    io_context: &mut IoContext<N>,
+    bit: usize,
+) -> IoResult<Vec<Rep3RingShare<PermRing>>> {
+    let len = inputs.len();
+    let mut bits = Vec::with_capacity(len);
+    for inp in inputs {
+        let a = inp.a.bit(bit as u64) as PermRing;
```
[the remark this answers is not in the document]
rw0x0: these ones are ok bit is a usize and the as PermRing casts from a bool. So both times we cast to larger stuff. But I get your point. At some point we should probably skim through the whole codebase to replace "as"

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -0,0 +1,625 @@
...
+        s = arithmetic::add(s, *f);
+        s1.push(s);
+    }
+    let mul1 = arithmetic::mul_vec(&f0, &s0, io_context)?;
+    let mul2 = arithmetic::mul_vec(&f1, &s1, io_context)?;
+    let perm = mul1
+        .into_iter()
+        .zip(mul2)
+        .map(|(a, b)| arithmetic::add(a, b))
+        .collect();
```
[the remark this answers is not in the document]
rw0x0: good observation! I replaced the two muls with local muls and reshare after the add. Saves a whole communication round, sending N elements, and less work during addition. Strictly better

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
rw0x0: have added a check in the beginning. We get an error when the size is too large and an instruction to increase ring size. Doubling ring size unfortuantely also doubles the communication, so I wanted to keep it restricted. I think sorting 4*10^9 elements is enough for now (in shared form the have 32 * 4* 10^9 * 2 bytes, i.e., 256 GB of data, so RAM will be high anyways)

rw0x0: I added a comment to the Perm ring as well

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
rw0x0: replaced

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
rw0x0: done

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
rw0x0: done

0xThemis (approved):

PR #297, opened 2024-12-18
feat: make check on zkey opt in with config value

fabian1409:


co-circom/co-circom/src/lib.rs
```diff
@@ -397,6 +397,9 @@ pub struct GenerateProofCli {
     #[arg(long)]
     #[serde(skip_serializing_if = "::std::option::Option::is_none")]
     pub zkey: Option<PathBuf>,
+    /// Perform checks on the zkey elements (can take a long time)
+    #[arg(long)]
+    pub check_zkey: bool,
```
dkales, line 402: Just for clarification, this implicitly is defaulted to false, and not mandatory to pass, right?

dkales (approved):

Issue #273, opened 2024-11-19
batch decompose ranges in co-noir


rw0x0:
Currently, decomposing the field elements for the range checks in co-noir is done by one garbled circuit for each decomposition. To reduce communication rounds, the decomposition can be batched. The mpc-core backend already has a garbled circuit implementation for that.

rw0x0:
Merged with #299

PR #299, opened 2024-12-23
Feat/batch decompose

florin5f:
do the decomposing in range checks in co-noir in batches instead of separately for shared values


rw0x0 (approved):

rw0x0:
Also includes the PR #294 and the PR #295. 
Closes issue #273

PR #312, opened 2025-01-31
feat: Add poseidon2 blackbox function support to co-Noir

rw0x0:
- Moves the Poseidon2 implementation of for the prover Transcript to the MPC core


mpc-core/src/protocols/rep3/arithmetic/types.rs
```diff
@@ -61,11 +61,19 @@ impl<F: PrimeField> Rep3PrimeFieldShare<F> {
     }
 
     /// Double the share in place
-    pub fn double(&mut self) {
+    pub fn double_in_place(&mut self) {
```
dkales, line 64: this make this a breaking change

florin5f (dismissed):

florin5f (approved):

PR #318, opened 2025-02-10
fix: Fix a bug with shifting BigUints in Range constraints

### LINKED ISSUE #317: `.assert_max_bit_size::<253>()` produces invalid verification key
I am getting different verification keys generated by `co-noir create-vk` and `bb write_vk_ultra_keccak_honk` if i am using `my_field.assert_max_bit_size::<253>()`. Funnily enough, bit sizes of 64, 128, 250 and 252 work fine(verifications keys are the same). Verification keys differ only if i use 253.

Proofs generated by `co-noir` and `bb` also differ.

Here is a repro. Check readme how to run:
https://github.com/olehmisar/co-noir-experiments/tree/om/assert_max_bit_size

rw0x0:
In our code we shifted: PrimeField::from(BigUint::one() << shift) which should have been a  PrimeField::from(uint256::one() << shift). Hence, when shift was > 256, our case producesd a huge value mod P, while the other case produced 0 mod P.

fixes #317


rw0x0:
Fixes #317

florin5f (approved):

PR #319, opened 2025-02-10
feat: optimize radix sort to take private and public inputs, such that public inputs do not have to be decomposed/bitinjected

### LINKED ISSUE #313: Optimize sorting in RAM/ROM for public values
We should think about optimizing the radix sort in ROM/RAM operations for coNoir when some values are public

rw0x0:
fixes #313


rw0x0:
Seems like I broke ROM/RAM in MPC with this change... Need to debug

florin5f (approved):

PR #320, opened 2025-02-11
feat: Add batched poseidon and merkle tree implementations

rw0x0:


mpc-core/src/gadgets/poseidon2/poseidon2_permutation.rs
```diff
@@ -241,6 +300,48 @@ mod test {
         }
     }
 
+    #[test]
+    fn posedon2_bn254_t2_kat1() {
```
dkales, line 304: could use a reference to where these come from

co-circom/co-circom/src/bin/poseidon2_bench.rs
```diff
@@ -109,24 +129,113 @@ fn main() -> color_eyre::Result<ExitCode> {
...
+    }
+
+    if config.merkle_size != 0 {
+        poseidon2_mt_benches::<F, T, D, ARITY, COMPRESSION_MODE>(config)?;
+    }
 
     Ok(ExitCode::SUCCESS)
 }
 
+#[allow(dead_code)]
```
dkales, line 181: why do these need `allow(dead_code)`?

rw0x0: because I may sometimes want to comment out some parts of the bench function without generating warnings :)

mpc-core/src/gadgets/poseidon2/poseidon2_permutation.rs
rw0x0: should all be compatible with the original poseidon2 generation script, will at a note

rw0x0: pushed comments to the kats and the parameter files themselves

dkales (approved):

dkales:
last commit is not conventional

PR #321, opened 2025-02-12
Update to Noir 1.0.0-beta.2 and barretenberg 0.72.1

rw0x0:


co-noir/co-noir/src/lib.rs
```diff
@@ -418,7 +419,7 @@ pub fn compute_circuit_size<P: Pairing>(
         constraint_system,
         recursive,
         0,
-        true,
+        1,
```
dkales, line 422: I guess this is the ultrahonk recursion flavor, can we make that an enum with repr u32 or whatever and the correct values assigned, since random `1`s here are pretty non-descriptive.

dkales (changes requested):

co-noir/co-noir/src/lib.rs
florin5f: changed it, but not with a repr u32 as I don't think we need it?

co-noir/co-builder/src/acir_format.rs
```diff
@@ -23,14 +23,20 @@ pub struct ProgramMetadata {
...
-    pub(crate) honk_recursion: u32, // honk_recursion means we will honk to recursively verify this
+    pub(crate) honk_recursion: HonkRecursion, // honk_recursion means we will honk to recursively verify this
     // circuit. This distinction is needed to not add the default
     // aggregation object when we're not using the honk RV.
     // 0 means we are not proving with honk
     // 1 means we are using the UltraHonk flavor
     // 2 means we are using the UltraRollupHonk flavor
     pub(crate) size_hint: usize,
 }
+#[derive(PartialEq)]
```
dkales, line 34: probably just add all of Debug, Clone, Copy, PartialEq, Eq, Hash by default

co-noir/co-builder/src/acir_format.rs
```diff
@@ -23,14 +23,20 @@ pub struct ProgramMetadata {
     // inside of another SNARK. For example, a recursive friendly proof may use Blake3Pedersen
     // for hashing in its transcript, while we still want a prove that uses Keccak for its
     // transcript in order to be able to verify SNARKs on Ethereum.
-    pub(crate) honk_recursion: u32, // honk_recursion means we will honk to recursively verify this
+    pub(crate) honk_recursion: HonkRecursion, // honk_recursion means we will honk to recursively verify this
     // circuit. This distinction is needed to not add the default
     // aggregation object when we're not using the honk RV.
     // 0 means we are not proving with honk
     // 1 means we are using the UltraHonk flavor
     // 2 means we are using the UltraRollupHonk flavor
```
dkales, line 31: this should also probably move down to the enum

co-noir/co-builder/src/acir_format.rs
florin5f: done

co-noir/co-builder/src/acir_format.rs
florin5f: done

dkales (approved):

florin5f:
dont merge yet, we found a little error in the transcript labels

florin5f (approved):

PR #333, opened 2025-02-24
feat: add plain zk prover and zk verifier

florin5f:


co-noir/ultrahonk/src/decider/prover.rs
```diff
@@ -73,26 +102,58 @@ impl<P: HonkCurve<TranscriptFieldType>, H: TranscriptHasher<TranscriptFieldType>
         circuit_size: u32,
         crs: &ProverCrs<P>,
         sumcheck_output: SumcheckOutput<P::ScalarField>,
+        has_zk: bool,
```
dkales, line 105: Maybe these could all be enums for more clarity?

rw0x0 (approved):

co-noir/ultrahonk/src/decider/prover.rs
rw0x0: updated

PR #340, opened 2025-03-10
feat: Add the MSM blackbox function to co-noir, which allows to use pedersen hash/commitment

rw0x0:


rw0x0:
Due to #350 I need to update some stuff here to be compatible again

rw0x0:
The updates for BB v0.82 (see #350) are now also included

mpc-core/src/protocols/rep3_ring/lut.rs
```diff
@@ -71,6 +71,62 @@ impl<N: Rep3Network> Rep3LookupTable<N> {
         Self::default()
     }
 
+    /// This is on optimized protocol that takes multiple pulbic LUTs and looks them up with the same index. It only creates the OHV once.
```
florin5f, line 74: *an, *public

florin5f (commented):
looks good to me, but maybe @dkales you can have another look at pointshare.rs and conversion.rs? (I guess these are also the ones where you had doubts @rw0x0?)
(also generator.rs scares me a bit, but I think it's fine)

rw0x0:
> looks good to me, but maybe @dkales you can have another look at pointshare.rs and conversion.rs? (I guess these are also the ones where you had doubts @rw0x0?) (also generator.rs scares me a bit, but I think it's fine)

Yes exactly, pointshare and conversion. Also in generators, most is copied/direct_translation of bb or noir code. @dkales please just have a look at the way I check for the correct type and my usage of OnceLock in the generators.rs file

mpc-core/src/protocols/rep3/detail.rs
```diff
@@ -408,3 +408,79 @@ fn low_depth_binary_sub_from_const<F: PrimeField, N: Rep3Network>(
     let res = binary::xor_public(&res, &BigUint::one(), io_context.id);
     Ok(res)
 }
+
+/// For curves of the form y^2 = x^3 + ax + b, computes the addition of two points.
+/// Note: This implementation assumes that at least one point is randomly chosen (as is e.g., the case for point_share_to_fieldshares). Thus, the special case that the x-coordinate of the two points are equal is only considered to be able to happen if the sum is infinity (as is the case when translating a share of the infinity point to fieldshares). Thus, we count the fact of the x coordinates being equal as infinity.
```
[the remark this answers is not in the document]
rw0x0: yes exactly, this function is used for the translation only

co-noir/co-acvm/src/mpc.rs
```diff
@@ -62,6 +64,13 @@ pub trait NoirWitnessExtensionProtocol<F: PrimeField> {
     /// Adds to acvm types. Both can either be public or shared
     fn add(&self, lhs: Self::AcvmType, rhs: Self::AcvmType) -> Self::AcvmType;
 
+    /// Adds to acvm points. Both can either be public or shared
```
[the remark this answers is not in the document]
rw0x0: done

mpc-core/src/protocols/rep3_ring/lut.rs
rw0x0: done

co-noir/co-builder/src/types/field_ct.rs
```diff
@@ -531,3 +1108,2063 @@ impl<P: Pairing, T: NoirWitnessExtensionProtocol<P::ScalarField>> WitnessCT<P, T
...
+        driver: &mut T,
+    ) -> std::io::Result<()> {
+        if self.is_constant() || self.skip_primality_test() {
+            return Ok(());
+        }
+        // if !self.is_constant() && !self.skip_primality_test() {
+        let cycle_group_modulus = if self.use_bn254_scalar_field_for_primality_test() {
+            BigUint::from(ark_bn254::Fr::MODULUS)
+        } else {
+            P::BaseField::MODULUS.into()
```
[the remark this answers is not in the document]
rw0x0: Yes, they have EmbeddedCurve::ScalarField, which is the scalarfield of 
using EmbeddedCurve = std::conditional_t<std::same_as<FF, bb::g1::coordinate_field>, curve::BN254, curve::Grumpkin>;

So yeah, they use Grumpkin:;ScalarField, which is our BN254 BaseField. Will add a comment

co-noir/co-builder/src/types/field_ct.rs
```diff
@@ -531,3 +1108,2063 @@ impl<P: Pairing, T: NoirWitnessExtensionProtocol<P::ScalarField>> WitnessCT<P, T
...
+
+    pub(crate) fn new_from_parts(x: P::ScalarField, y: P::ScalarField, is_infinity: bool) -> Self {
+        if is_infinity {
+            return Self::default();
+        }
+        let x = FieldCT::from(x);
+        let y = FieldCT::from(y);
+        let is_infinity = BoolCT::from(false);
+
+        // TACEO TODO: We should assert a valid point here
```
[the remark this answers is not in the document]
rw0x0: This is grumpkin, but at this point this is behind a trait, so we would need to make this if else structure here where we check for type since P: Pairing is not really generous in exposing the required stuff. Not sure if we want to do that though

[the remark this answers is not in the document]
rw0x0: The reason why I made this commetn is that checking if the value in on the curve is an assert (i.e., sanity check) in BB

co-noir/co-builder/src/types/generators.rs
```diff
@@ -0,0 +1,344 @@
...
+pub(crate) fn default_generators<C: CurveGroup>() -> &'static [C::Affine; NUM_DEFAULT_GENERATORS] {
+    if TypeId::of::<C>() == TypeId::of::<ark_grumpkin::Projective>() {
+        let gens = _default_generators_grumpkin();
+        // Safety: We checked that the types match
+        unsafe {
+            std::mem::transmute::<
+                &[ark_grumpkin::Affine; NUM_DEFAULT_GENERATORS],
+                &[C::Affine; NUM_DEFAULT_GENERATORS],
+            >(gens)
+        }
```
dkales, line 19: in general, these checks, especially if we working with references, can be written with completely safe rust and producing the same machine code using something like: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d0e1de8b9c6116521feca4fb1aa66037

I think we can rewrite most of them to this easily, without unsafe transmutes (even though they are fine here.

co-noir/co-builder/src/types/generators.rs
```diff
@@ -0,0 +1,344 @@
...
+        } else {
+            point
+        }
+    } else {
+        _hash_to_curve_bn254(seed, attempt_count + 1)
+    }
+}
+//
+pub(crate) fn generate_fixed_base_tables<C: CurveGroup>(
+) -> &'static [Vec<Vec<C::Affine>>; FixedBaseParams::NUM_FIXED_BASE_MULTI_TABLES] {
```
dkales, line 198: how large are these (mem usage)?

mpc-core/src/protocols/rep3/conversion.rs
```diff
@@ -604,3 +605,201 @@ pub fn b2y2a_streaming<F: PrimeField, N: Rep3Network>(
     let y = b2y(x, delta, io_context)?;
     y2a_streaming(y, delta, io_context)
 }
+
+#[expect(clippy::type_complexity)]
+pub(crate) fn point_share_to_fieldshares_pre<C: CurveGroup, N: Rep3Network>(
```
dkales, line 610: think these should also have doc comments

mpc-core/src/protocols/rep3/detail.rs
```diff
@@ -387,3 +389,79 @@ fn low_depth_binary_sub_from_const<F: PrimeField, N: Rep3Network>(
...
+                *(&x as *const C::BaseField as *const ark_grumpkin::Fq),
+                *(&y as *const C::BaseField as *const ark_grumpkin::Fq),
+            )
+        };
+        let result: ark_grumpkin::Projective = ark_grumpkin::Affine::new(x, y).into();
+        unsafe { *(&result as *const ark_grumpkin::Projective as *const C) }
+    } else {
+        panic!("Unsupported curve {}", std::any::type_name::<C>());
+    };
+    Ok(point)
```
dkales, line 466: same here for the any downcast, can probably move this to safe code without performance losses

co-noir/co-builder/src/types/generators.rs
rw0x0: I am struggling with the slices, it works on fixed types though

mpc-core/src/protocols/rep3/detail.rs
rw0x0: I removed the unsafe wherever I have no slices

co-noir/co-builder/src/types/generators.rs
rw0x0: The inner most vec is 512, the outer Vec is 14, 14, 15, 15. So in total this means 512*58 = 29696, which is 59392 field elements, which corresponds to roughly 2MB

rw0x0: BB also keeps this table in memory

dkales (approved):

PR #345, opened 2025-03-18
feat!: acvm now can store values and returns the output of circuit

0xThemis:
This came up during the visit in London. We somehow need to persist the private shared state that we create for the MPC nodes. For that we added a foreign_function hook that stores shares of something to then later retrieve them by the caller.

When I were at it, I also added a convenient way to retrieve the output of the circuit.


co-noir/co-brillig/src/taceo_std_lib.rs
```diff
@@ -0,0 +1,58 @@
...
+                "Invalid signature for TACEO store. Second paramter must be array of field elements."
+            );
+        };
+        if self
+            .persistent_shared_state
+            .insert(identifer.clone(), serialized)
+            .is_some()
+        {
+            eyre::bail!(format!("duplicate entry for shared state id: {identifer}"));
+        }
```
dkales, line 54: is this really  the intended behavior that you cannot overwrite stuff ever in a program?

co-noir/co-brillig/src/brillig_vm.rs
```diff
@@ -408,6 +461,7 @@ where
             driver: driver1,
             calldata: self.calldata.clone(),
             unconstrained_functions: self.unconstrained_functions.clone(),
+            persistent_shared_state: self.persistent_shared_state.clone(),
```
dkales, line 464: cloning here seems wrong since we cannot put stuff into the PSS in forks anyways? Maybe just an `Arc` or `Arc<Mutex>` would fit better here? Just something to think about, can leave it at clone

co-noir/co-acvm/src/value_store.rs
```diff
@@ -0,0 +1,61 @@
+use std::collections::HashMap;
+
+use ark_ff::PrimeField;
+use co_brillig::mpc::BrilligDriver;
+
+use crate::mpc::NoirWitnessExtensionProtocol;
+
+/// TODO better name?
```
dkales, line 8: `PersistentValueStore`?

dkales: Even though its not really persistent yet I guess

co-noir/co-acvm/src/solver.rs
```diff
@@ -376,12 +393,21 @@ where
             }
         }
         tracing::trace!("we are done! Opening results...");
-        self.open_results(&functions[self.function_index])?;
+        let output = self.open_results(&functions[self.function_index])?;
+        self.value_store.set_output(output);
         tracing::trace!("Done! Wrap things up.");
+
+        // TODO here we need to store the output
```
dkales, line 400: do we leave it like that ATM?

co-noir/co-brillig/src/brillig_vm.rs
```diff
@@ -40,7 +45,7 @@ where
 {
     /// Indicates that the run of the Brillig-VM was a sucess. Holds
     /// the computed values
-    Success(Vec<T::BrilligType>),
+    Success(Vec<T::BrilligType>, HashMap<String, Vec<T::BrilligType>>),
```
dkales, line 48: this could be a struct variant with descriptive names

co-noir/co-brillig/src/brillig_vm.rs
```diff
@@ -315,7 +342,7 @@ where
                     let result = self.driver.cmux(condition.clone(), truthy, falsy)?;
                     final_result.push(result);
                 }
-                Ok(Some(CoBrilligResult::Success(final_result)))
+                Ok(Some(CoBrilligResult::Success(final_result, HashMap::new())))
```
dkales, line 345: this does not seem right? can the PSS not have some values before the if and would they not get dropped here?

co-noir/co-brillig/src/taceo_std_lib.rs
0xThemis: This is something we need to discuss you are right - I went with this because we intended that this method will be called automatically when calling:
`PrivateState::new() `
in Noir. Then when loading the PSS we call 
`PrivateState::from_data_and_commitment(state, state_r, state_c)`

so this method will not be called again.

co-noir/co-brillig/src/brillig_vm.rs
0xThemis: I'll leave it for the moment, because the brillig fork is not supported for that anyways and I think we need to rewrite the forking anyways in the forseeable future.

co-noir/co-acvm/src/value_store.rs
0xThemis: I'll go with PssStore?

co-noir/co-acvm/src/pss_store.rs
```diff
@@ -6,7 +6,7 @@ use co_brillig::mpc::BrilligDriver;
 use crate::mpc::NoirWitnessExtensionProtocol;
 
 /// TODO better name?
```
dkales, line 8: remove todo?

dkales (dismissed):

dkales (approved):

PR #349, opened 2025-03-31
Reduce work for Groth16 REP3 by working over un-replicated shares as much as possible

dkales:
- **wip on groth16 improvements**
- **move rs to local mul**
- **feat: multi threaded eval constraint in groth16**
- **Revert "feat: multi threaded eval constraint in groth16"**

### DIFF
--- a/co-circom/co-groth16/src/groth16.rs
+++ b/co-circom/co-groth16/src/groth16.rs
@@ -118,14 +118,16 @@ where
             )
         }
 
-        let private_witness = Arc::new(private_witness.witness);
-        let h = self.witness_map_from_matrices(&zkey, &public_inputs, &private_witness)?;
+        let h = self.witness_map_from_matrices(&zkey, &public_inputs, &private_witness.witness)?;
         let (r, s) = (self.driver.rand()?, self.driver.rand()?);
 
-        let private_witness = Arc::into_inner(private_witness).unwrap();
-        let private_witness_half_share =
-            private_witness.into_iter().map(T::to_half_share).collect();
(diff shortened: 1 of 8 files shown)


co-circom/co-groth16/src/groth16.rs
```diff
@@ -123,6 +122,11 @@ where
         let h = self.witness_map_from_matrices(&zkey, &public_inputs, &private_witness)?;
         let (r, s) = (self.driver.rand()?, self.driver.rand()?);
 
+        let private_witness = Arc::into_inner(private_witness).unwrap();
```
rw0x0, line 125: Here we consume an Arc that we have created 5 lines before, but we never used the Arc, is this necessary?

co-circom/co-groth16/src/mpc/shamir.rs
```diff
@@ -176,25 +179,58 @@ impl<P: Pairing, N: ShamirNetwork> CircomGroth16Prover<P>
...
+            });
+            let opened = core::reconstruct_point(
+                &opened.join().expect("can join")?,
+                &self.protocol0.open_lagrange_2t,
+            );
+            Ok((opened, mul_result.join().expect("can join")?))
         })
     }
+
+    fn to_half_share(a: Self::ArithmeticShare) -> <P as Pairing>::ScalarField {
```
rw0x0, line 216: Maybe we can add a comment that a degree-t share is also a valid degree-2t share, otherwise this might be confusing with the comment that a halfshare is always a degree-2t share

rw0x0 (approved):

PR #357, opened 2025-04-03
feat: Add missing int_div and mod code in the circom rep3 witext backend

rw0x0:


mpc-core/src/protocols/rep3/yao.rs
```diff
@@ -695,6 +695,18 @@ pub fn slice_arithmetic<F: PrimeField, N: Rep3Network>(
     slice_arithmetic_many(&[input], io_context, msb, lsb, bitsize)
 }
 
+/// Computes input % 2^divisor_bit.
+pub fn share_mod_power_2<F: PrimeField, N: Rep3Network>(
```
dkales, line 699: naming is usually `field_` here, no?

rw0x0: adapted

PR #364, opened 2025-04-14
Feat/bb blake3

florin5f:


florin5f:
do not review this yet, will clean it up once blake2 is merged into main

PR #374, opened 2025-05-03
co-noir: make AcirFormat implement Clone

teddav:


teddav:
i was told by @fabian1409 that this is not necessary.
So i guess this can be closed

PR #383, opened 2025-06-18
feat: added calls so that mpc-vm doesnt consume compiled circuit

0xThemis:


co-circom/circom-mpc-vm/src/types.rs
```diff
@@ -253,14 +253,57 @@ impl<F: PrimeField> CoCircomCompilerParsed<F> {
...
             MpcAccelerator::from_config(MpcAcceleratorConfig::from_env()),
             vm_config,
             batch_size,
         )
     }
 
+    /// Consumes a [`NetworkConfig`], and constructs an instance of [`Rep3WitnessExtension`].
+    ///
+    /// # Arguments
+    /// - `network_config`: A network configuration specifying how to connect to the other two parties.
```
fabian1409, line 267: nit: vm_config missing

also do we need these two functions? them VM is never constructed in user facing code afaik, so we already have a wrapper [here](https://github.com/TaceoLabs/co-snarks/blob/main/co-circom/co-circom/src/lib.rs#L315C8-L315C29)

fabian1409 (approved):

PR #384, opened 2025-06-23
refactor: unify MPC networks, split protocol state and networking to allow fork of state without network, replace io::Error with eyre

fabian1409:


co-circom/circom-mpc-vm/src/mpc/batched_rep3.rs
```diff
@@ -75,8 +82,8 @@ impl<F: PrimeField> From<Vec<ArithmeticShare<F>>> for BatchedRep3VmType<F> {
 }
 
 #[expect(unused_variables)]
-impl<F: PrimeField, N: Rep3Network> VmCircomWitnessExtension<F>
-    for BatchedCircomRep3VmWitnessExtension<F, N>
+impl<F: PrimeField, N: Network> VmCircomWitnessExtension<F>
```
0xThemis, line 87: nit: the name "network" may be kinda ambigous

co-circom/circom-mpc-vm/src/mpc/batched_rep3.rs
```diff
@@ -311,13 +316,12 @@ impl<F: PrimeField, N: Rep3Network> VmCircomWitnessExtension<F>
 
     fn compare_vm_config(&mut self, config: &crate::mpc_vm::VMConfig) -> eyre::Result<()> {
         let ser = bincode::serialize(&config)?;
-        self.io_context0.network.send_next(ser)?;
-        let rcv: Vec<u8> = self.io_context0.network.recv_prev()?;
+        network::send_next(self.net0, ser)?;
```
0xThemis, line 321: nit: also network module is somewhat ambigous

co-circom/circom-mpc-vm/src/mpc/rep3.rs
```diff
@@ -58,29 +50,29 @@ impl<F: PrimeField> Default for Rep3VmType<F> {
...
     plain: CircomPlainVmWitnessExtension<F>,
 }
 
-impl<F: PrimeField, N: Rep3Network> CircomRep3VmWitnessExtension<F, N> {
-    pub fn from_network(network: N, a2b_type: conversion::A2BType) -> io::Result<Self> {
-        let mut io_context = IoContext::init(network)?;
-        io_context.set_a2b_type(a2b_type);
-        let io_context_fork = io_context.fork()?;
+impl<'a, F: PrimeField, N: Network> CircomRep3VmWitnessExtension<'a, F, N> {
+    pub fn new(net0: &'a N, net1: &'a N) -> eyre::Result<Self> {
```
0xThemis: we need to be able to set conversion type

mpc-net/src/lib.rs
```diff
@@ -1,79 +1,82 @@
...
-    /// The wrapped network handler
-    pub inner: MpcNetworkHandler,
-}
-
-impl MpcNetworkHandlerWrapper {
-    /// Create a new wrapper
-    pub fn new(runtime: Runtime, inner: MpcNetworkHandler) -> Self {
-        Self { runtime, inner }
-    }
+/// A MPC network that can be used to send and receive data to and from other patires
```
0xThemis: typo "patires"

mpc-net/src/lib.rs
```diff
@@ -184,42 +192,168 @@ impl MpcNetworkHandler {
...
-            my_id: config.my_id,
+        Ok((connections, endpoints))
+    }
+
+    #[expect(clippy::complexity)]
+    pub fn get_streams(
+        &self,
+    ) -> eyre::Result<(
+        IntMap<usize, tokio::sync::mpsc::Sender<Vec<u8>>>,
+        IntMap<usize, Mutex<tokio::sync::mpsc::Receiver<Vec<u8>>>>,
```
0xThemis: Why must the Receiver be in a Mutex?

mpc-net/src/lib.rs
```diff
@@ -184,42 +192,168 @@ impl MpcNetworkHandler {
...
+
+            let mut write = FramedWrite::new(send_stream, codec.clone());
+            let mut read = FramedRead::new(recv_stream, codec);
+
+            let (send_tx, mut send_rx) = tokio::sync::mpsc::channel(32);
+            let (recv_tx, recv_rx) = tokio::sync::mpsc::channel(32);
+
+            self.rt.spawn(async move {
+                while let Some(frame) = send_rx.recv().await {
+                    write.send(Bytes::from(frame)).await?;
```
0xThemis: do we catch that error somewhere? Dont know if this is a problem but we propagate the error silently I think. The task then dies, therefore the next write will fail with channel dropped. 
Maybe it would be nice to log the error and then break from the task so that we can investigate the error (same for below)

mpc-net/src/lib.rs
```diff
@@ -184,42 +192,168 @@ impl MpcNetworkHandler {
...
+        for (id, conn) in self.connections.iter() {
+            if self.id < *id {
+                let mut send = conn.open_uni().await?;
+                send.write_all(b"done").await?;
+            } else {
+                let mut recv = conn.accept_uni().await?;
+                let mut buffer = vec![0u8; b"done".len()];
+                recv.read_exact(&mut buffer).await.map_err(|_| {
+                    std::io::Error::new(std::io::ErrorKind::BrokenPipe, "failed to recv done msg")
+                })?;
```
0xThemis: if we have broken pipe to first connection we cannot call function to second one. Maybe this leads to a problem for peer?

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
-                send_stream.write_u32(self.my_id as u32).await?;
-                let conn = Channel::new(recv_stream, send_stream, codec.clone());
-                assert!(channels.insert(id, conn).is_none());
+impl Network for QuicNetwork {
+    fn id(&self) -> usize {
+        self.id
+    }
+
+    fn send(&self, to: usize, data: &[u8]) -> eyre::Result<()> {
+        let stream = self.send.get(to).context("while get stream in send")?;
```
0xThemis: maybe have more meaningful context message like:
.context("party id out-of-bounds") or similar

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+                        stream.write_u64::<BigEndian>(id as u64)?;
+                        nets[i]
+                            .send
+                            .insert(other_id, Mutex::new(stream.try_clone().unwrap()));
+                        let (tx, rx) = mpsc::channel();
+                        std::thread::spawn(move || {
+                            loop {
+                                let len = stream.read_u32::<BigEndian>()? as usize;
+                                let mut data = vec![0; len];
+                                stream.read_exact(&mut data)?;
```
0xThemis: Also here we silently drop the error because no one is listening to the thread handle write? Logically it is ok because we drop the channel. Nevertheless, logging any errors solves some headaches in the future I think

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+                            }
+                            std::thread::sleep(Duration::from_millis(50));
+                        };
+                        stream.set_write_timeout(Some(timeout))?;
+                        stream.set_nodelay(true)?;
+                        stream.write_u64::<BigEndian>(i as u64)?;
+                        stream.write_u64::<BigEndian>(id as u64)?;
+                        nets[i]
+                            .send
+                            .insert(other_id, Mutex::new(stream.try_clone().unwrap()));
```
0xThemis: expect instead of unwrap with reasonable message.
Also is this ok even to unwrap?

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+                        let (tx, rx) = mpsc::channel();
+                        std::thread::spawn(move || {
+                            loop {
+                                let len = stream.read_u32::<BigEndian>()? as usize;
```
0xThemis: As discussed async - make 5 bytes. Maybe simply doing u64 is enough anyways - up to you to decide. Maybe we can do a small benchmark and if it is not measurable just stick with u64.

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+                        stream.set_nodelay(true)?;
+                        let i = stream.read_u64::<BigEndian>()? as usize;
+                        let other_id = stream.read_u64::<BigEndian>()? as usize;
+                        nets[i]
+                            .send
+                            .insert(other_id, Mutex::new(stream.try_clone().unwrap()));
+                        let (tx, rx) = mpsc::channel();
+                        std::thread::spawn(move || {
+                            loop {
+                                let len = stream.read_u32::<BigEndian>()? as usize;
```
0xThemis: same as above

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+                                let mut data = vec![0; len];
+                                stream.read_exact(&mut data)?;
+                                tx.send(data)?;
+                            }
+                            #[allow(unreachable_code)]
```
0xThemis: expect instead of allow?
also if we dont error and instead log and break we dont need the eyre::ok (true for all such tasks)

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
-                let mut buffer = vec![0u8; b"done".len()];
-                recv.read_exact(&mut buffer).await.map_err(|_| {
-                    std::io::Error::new(std::io::ErrorKind::BrokenPipe, "failed to recv done msg")
-                })?;
+    fn send(&self, to: usize, data: &[u8]) -> eyre::Result<()> {
+        let mut stream = self
+            .send
+            .get(to)
+            .context("while get stream in send")?
+            .lock();
```
0xThemis: talked about lock async (can potentially remove it, but maybe not needed)

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+                            stream.write_u64::<BigEndian>(id as u64)?;
+                            stream.write_u8(s)?;
+
+                            if s == STREAM_0 {
+                                nets[i]
+                                    .send
+                                    .insert(other_id, Mutex::new(TlsStream::Client(stream)));
+                            } else {
+                                let (tx, rx) = mpsc::channel();
+                                std::thread::spawn(move || {
```
0xThemis: same remarks as above for tasks - also here we have an unwrap. We want tracing::warn and break to kill task

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+                            let mut stream = StreamOwned::new(conn, stream);
+
+                            let i = stream.read_u64::<BigEndian>()? as usize;
+                            let other_id = stream.read_u64::<BigEndian>()? as usize;
+                            let s_ = stream.read_u8()?;
+
+                            if s_ == STREAM_0 {
+                                let (tx, rx) = mpsc::channel();
+                                std::thread::spawn(move || {
+                                    loop {
```
0xThemis: same as above tasks

mpc-net/src/lib.rs
```diff
@@ -229,95 +363,445 @@ impl MpcNetworkHandler {
...
+            .get(from)
+            .context("while get stream in recv")?
+            .lock();
+        Ok(queue.recv_timeout(self.timeout)?)
+    }
+}
+
+/// A MPC network using [mpsc::channel]s. Used for testing.
+#[derive(Debug)]
+pub struct TestNetwork {
```
0xThemis: Generally, I think the networks should be in dedicated files/modules. This file is way too long (this is just my opinion, but I think I will be annoyed searching something in this file). 

This specific Network should be in a dedicated module with cfg(test). If not possible then at least feature gated.

co-circom/circom-mpc-vm/src/mpc/rep3.rs
```diff
@@ -582,13 +574,12 @@ impl<F: PrimeField, N: Rep3Network> VmCircomWitnessExtension<F>
 
     fn compare_vm_config(&mut self, config: &VMConfig) -> eyre::Result<()> {
         let ser = bincode::serialize(&config)?;
-        self.io_context0.network.send_next(ser)?;
-        let rcv: Vec<u8> = self.io_context0.network.recv_prev()?;
+        network::send_next(self.net0, ser)?;
```
0xThemis, line 578: Just an idea, I said that network is ambigous. Maybe we can remove network as module and just have mpc_net. Then we can write mpc_net::send_next which is not ambigous anymore.

co-circom/circom-mpc-vm/src/mpc/rep3.rs
```diff
@@ -582,13 +574,12 @@ impl<F: PrimeField, N: Rep3Network> VmCircomWitnessExtension<F>
...
+        let rcv: Vec<u8> = network::recv_prev(self.net0)?;
```
0xThemis: nit: it triggers me that there is rcv = recv.prev and not recv = recv.prev

co-circom/circom-mpc-vm/src/mpc_vm.rs
```diff
@@ -1142,7 +1143,7 @@ impl<F: PrimeField, N: Rep3Network> Rep3WitnessExtension<F, N> {
                 parser.fun_decls.clone(),
                 parser.templ_decls.clone(),
                 parser.string_table.clone(),
-                mpc_accelerator,
+                MpcAccelerator::from_config(MpcAcceleratorConfig::from_env()),
```
0xThemis, line 1145: why cant we provide the accelerator anymore?

co-circom/circom-mpc-vm/src/mpc_vm.rs
```diff
@@ -1217,11 +1181,11 @@ impl<F: PrimeField, N: Rep3Network> BatchedRep3WitnessExtension<F, N> {
             main: parser.main.clone(),
             ctx: WitnessExtensionCtx::new(
                 signals,
-                constant_table,
+                batched_constant_table,
                 parser.fun_decls.clone(),
                 parser.templ_decls.clone(),
                 parser.string_table.clone(),
-                mpc_accelerator,
+                MpcAccelerator::from_config(MpcAcceleratorConfig::from_env()),
```
0xThemis, line 1187: Same as above - why no provided acc anymore?

co-circom/co-groth16/src/groth16.rs
```diff
@@ -299,66 +308,73 @@ impl<P: Pairing, T: CircomGroth16Prover<P>> CoGroth16<P, T> {
...
-        // pairs)
-        let protocol1 = protocol0.fork_with_pairs(2)?;
-        let driver = ShamirGroth16Driver::new(protocol0, protocol1);
-        let prover = CoGroth16 {
-            driver,
-            phantom_data: PhantomData,
-        };
+        let preprocessing = ShamirPreprocessing::new(num_parties, threshold, num_pairs, net0)?;
+        let mut state0 = ShamirState::from(preprocessing);
+        // let mut state1 = state0.fork(1)?;
```
0xThemis: Remove commend line

co-circom/co-groth16/src/groth16.rs
```diff
@@ -299,66 +308,73 @@ impl<P: Pairing, T: CircomGroth16Prover<P>> CoGroth16<P, T> {
...
+        let mut state1 = state0.fork(0)?;
```
0xThemis: why 0 here?

co-noir/co-acvm/src/mpc/shamir.rs
```diff
@@ -145,20 +144,20 @@ impl<F: PrimeField> From<ShamirBrilligType<F>> for ShamirAcvmType<F> {
...
 
-    type BrilligDriver = ShamirBrilligDriver<F, N>;
+    type BrilligDriver = ShamirBrilligDriver<'a, F, N>;
 
-    fn init_brillig_driver(&mut self) -> std::io::Result<Self::BrilligDriver> {
-        Ok(ShamirBrilligDriver::with_protocol(
-            self.protocol.fork_with_pairs(0)?, // TODO maybe have some pairs here
-        ))
+    fn init_brillig_driver(&mut self) -> eyre::Result<Self::BrilligDriver> {
+        // TODO we just copy the net ref here this is not safe if used concurrently
```
0xThemis, line 159: do we need to address this TODOs?

mpc-core/src/lib.rs
```diff
@@ -11,9 +11,34 @@ pub mod protocols;
...
+/// Trait for MPC protocol states
+pub trait MpcState: Sized {
+    /// The type of a party id
+    type PartyID: Clone + Copy + Send + Sync;
+
+    /// Get the id of the party
+    fn id(&self) -> Self::PartyID;
+
+    // TODO maybe use fork() and fork_with(n: usize)
+    /// Crate a new state from self
```
0xThemis, line 28: typo: Crate instead of Create

mpc-core/src/protocols/rep3.rs
```diff
@@ -13,8 +13,108 @@ pub mod poly;
...
+    pub rngs: Rep3CorrelatedRng,
+    /// The rng type
+    pub rng: RngType,
+    /// A2B conversion type
+    pub a2b_type: A2BType,
+}
+
+impl Rep3State {
+    /// Create a new [Rep3State]
+    pub fn new<N: Network>(net: &N) -> eyre::Result<Self> {
```
0xThemis: we need to be able to set a2b type

mpc-core/src/protocols/shamir/network.rs
```diff
@@ -1,317 +1,320 @@
...
-    }
-
-    fn recv_many<F: CanonicalDeserialize>(&mut self, from: usize) -> std::io::Result<Vec<F>> {
-        let data = self.recv_bytes(from)?;
-
-        let res = Vec::<F>::deserialize_uncompressed_unchecked(&data[..])
-            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
-
-        Ok(res)
+// TODO use rayon
```
0xThemis: address this todo?

mpc-core/src/protocols/shamir/network.rs
```diff
@@ -1,317 +1,320 @@
...
-            for other_id in 0..num_parties {
-                if other_id != id {
-                    let chan = channels.remove(&other_id).expect("to find channel");
-                    channels_.insert(other_id, ChannelHandle::manage(chan));
-                }
-            }
-
-            if !channels.is_empty() {
-                panic!("unexpected channels found")
+// TODO use rayon
```
0xThemis: address this todo?

mpc-net/src/lib.rs
fabian1409: solved using crossbeam_channel

mpc-net/src/lib.rs
fabian1409: afaik this only fails if the creation of a new file descriptor failes, so i think is safe to expect

mpc-net/src/tcp.rs
```diff
@@ -0,0 +1,223 @@
...
+        let domain = match bind_addr {
+            SocketAddr::V4(_) => Domain::IPV4,
+            SocketAddr::V6(_) => Domain::IPV6,
+        };
+        let socket = Socket::new(domain, Type::STREAM, None)?;
+        socket.set_reuse_address(true)?;
+        if bind_addr.is_ipv6() {
+            socket.set_only_v6(false)?;
+        }
+        socket.set_read_timeout(Some(timeout))?;
```
dkales, line 117: we might also want to add a keepalive config to the sockets once they are established, like we did for quinn?

dkales: same for tls

mpc-net/src/lib.rs
```diff
@@ -1,323 +1,42 @@
...
-    /// Create a new wrapper
-    pub fn new(runtime: Runtime, inner: MpcNetworkHandler) -> Self {
-        Self { runtime, inner }
-    }
+#[cfg(feature = "quic")]
+pub mod quic;
+#[cfg(feature = "tcp")]
+pub mod tcp;
+#[cfg(feature = "thread")]
+pub mod thread;
```
dkales: why is this called thread and not `local` or `test`?

mpc-net/examples/three_party.rs
dkales, line 1: can we also have an example for quic and tls (no need for config files, can just create the configs inline as well.)

mpc-net/src/quic.rs
```diff
@@ -0,0 +1,472 @@
...
+            id,
+            send,
+            recv,
+            conn_handler,
+            timeout,
+        })
+    }
+
+    /// Prints the connection statistics.
+    pub fn print_connection_stats(&self, out: &mut impl std::io::Write) -> std::io::Result<()> {
```
dkales, line 470: can we have this tracking for tcp/tls as well?

mpc-net/src/tcp.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    fn recv(&self, from: usize) -> eyre::Result<Vec<u8>> {
+        let queue = self.recv.get(from).context("party id out-of-bounds")?;
+        Ok(queue.recv_timeout(self.timeout)?)
+    }
+}
+
+fn read_next_frame(stream: &mut TcpStream) -> std::io::Result<Vec<u8>> {
+    let len = stream.read_u64::<BigEndian>()? as usize;
```
dkales, line 251: might be nice to have a configurable max length here such that other parties cannot do DOS attacks.
In general I think that we should not sent to large frames here if possible...

mpc-core/src/protocols/rep3/arithmetic.rs
```diff
@@ -227,85 +231,79 @@ pub fn inv<F: PrimeField, N: Rep3Network>(
...
         .iter()
         .map(|share| (share.a, share.b))
         .collect::<(Vec<F>, Vec<F>)>();
-    let c = io_context.network.reshare_many(&b)?;
+    let c = network::reshare_many(net, &b)?;
     Ok(izip!(a, b, c).map(|(a, b, c)| a + b + c).collect_vec())
 }
 
 /// Computes a CMUX: If cond is 1, returns truthy, otherwise returns falsy.
 /// Implementations should not overwrite this method.
```
dkales: This comment seems no longer relevant?

co-circom/co-circom/src/bin/co-circom.rs
```diff
@@ -835,12 +831,10 @@ where
     // Extend the witness
     tracing::info!("Starting witness generation...");
     let start = Instant::now();
-    let (result_witness_share, mpc_net) =
-        co_circom::generate_witness_rep3::<P>(circuit, input_share, mpc_net, config.vm)?;
+    let result_witness_share =
+        co_circom::generate_witness_rep3::<P, _>(&circuit, input_share, config.vm, &net0, &net1)?;
```
dkales, line 829: I don't like that this now needs to take 2 networks. Let's leave it for now but I still feel like some form of session multiplexing to allow for forks would be nice.

co-circom/co-circom/src/bin/co-circom.rs
```diff
@@ -14,7 +14,7 @@ use figment::{
     Figment,
     providers::{Env, Format, Serialized, Toml},
 };
-use mpc_net::config::NetworkConfigFile;
+use mpc_net::tcp::{NetworkConfig, TcpNetwork};
```
dkales, line 17: should the co-circom binary use the tls network by default?

co-circom/co-groth16/src/groth16.rs
```diff
@@ -107,28 +109,23 @@ fn root_of_unity_for_groth16<F: PrimeField + FftField>(
 
 /// A Groth16 proof protocol that uses a collaborative MPC protocol to generate the proof.
 pub struct CoGroth16<P: Pairing, T: CircomGroth16Prover<P>> {
-    pub(crate) driver: T,
-    phantom_data: PhantomData<P>,
+    phantom_data0: PhantomData<P>,
+    phantom_data1: PhantomData<T>,
```
dkales: can just use a single PhantomData<(P,T)> if you want.

co-circom/co-groth16/src/groth16.rs
```diff
@@ -299,66 +309,72 @@ impl<P: Pairing, T: CircomGroth16Prover<P>> CoGroth16<P, T> {
         g_c += h_acc;
 
         let g2_b = s_g2;
-        let (g_c_opened, g2_b_opened) = self.driver.open_two_half_points(g_c, g2_b)?;
+        let (g_c_opened, g2_b_opened) = rayon::join(
+            || T::open_half_point(g_c, net0, state0),
```
dkales, line 312: Maybe make open_half_points take a vector and use that variant instead of this join here?

dkales: and we can still use the vector variant above.

co-circom/co-plonk/src/lib.rs
```diff
@@ -125,25 +129,60 @@ mod plonk_utils {
...
+                    rayon::join($f2, || {
+                        rayon::join($f3, || {
+                            rayon::join($f4, || rayon::join($f5, || rayon::join($f6, $f7)))
+                        })
+                    })
+                })
+            });
+            (r0, r1, r2, r3, r4, r5, r6, r7)
         }};
     }
```
dkales, line 168: Is there a benefit in nesting these in a tree-like fashion?

co-circom/co-plonk/src/plonk.rs
```diff
@@ -272,29 +259,6 @@ where
     }
 }
```
dkales, line 261: Did we not use this anywhere?

release-please-config.json
```diff
@@ -112,14 +112,6 @@
       "draft": false,
       "prerelease": false
     },
```
dkales, line 114: do not forget the second release plase file

mpc-net/src/lib.rs
fabian1409: Franco didnt like the module name "test" so i changed it
gonna change it to "local" and "LocalNetwork"

mpc-net/src/tcp.rs
fabian1409: there are write/read_u48 methods, do you think 1TB is fine? with u32 we are limited to 4GB which was to little in the past.

could also use a simple varint solution where we sent 1 byte that defines the byte_len of the data_len and the the data_len

mpc-core/src/protocols/rep3/arithmetic.rs
fabian1409: probably true yes, i'll remove it

co-circom/co-circom/src/bin/co-circom.rs
fabian1409: yea same, but since dynamic "fork" of anything other than quic is probably slow, i dont see any other option right now.

we can of course use the approach we designed for the "mpc-engine" in the future which should make this less ugly

i made this changes with a possible use of our "mpc-engine" in mind, so it shouldn't be to much work to add it in the future. then all these functions would take &MpcEngine and get however many networks they need from there

co-circom/co-circom/src/bin/co-circom.rs
fabian1409: idk maybe? i chose tcp because then one could use it for testing without care about certificates
but its a easy to replace so can do that if you prefer the tls one

co-circom/co-plonk/src/lib.rs
fabian1409: i dont think so, is there one for nesting it evenly?

co-circom/co-plonk/src/plonk.rs
fabian1409: just moved it to lib.rs so its next to Rep3 and Shamir prove impl same as for Groth16

co-circom/co-groth16/src/groth16.rs
fabian1409: 1 point is G1 and the other G2 so for a array i would have to serialize them before which is kinda annoying imo
i re-add the dedicated open_two_half_points function we had before for this case

fabian1409: ok the old function did the same thing internally because we pass data for the network functions that is ArkSerialize and that doesnt really work with different types e.g. tuple (P::G1, P::G2)

i would need to write a new function for broadcast that takes bytes as data then i can do this with 1 network
idk if that is cleaner than using 2 networks (since we already have 2 and did it this way before)

mpc-net/src/tcp.rs
dkales: I think even writes with 4 GB should not happen TBH. We should think about this at some point
Let's leave it as is for now

fabian1409: added config option to limit frame len

dkales (approved):

PR #391, opened 2025-07-03
ci: add GITHUB_TOKEN env to CI to allow more downloads via github API

fabian1409:
not sure if [release.yml](https://github.com/TaceoLabs/co-snarks/blob/main/.github/workflows/release.yml) also needs the token?

### DIFF
--- a/.github/workflows/rolling.yml
+++ b/.github/workflows/rolling.yml
@@ -30,6 +30,8 @@ jobs:
         run: cargo generate-lockfile
       - name: cargo test --locked
         run: cargo test --profile=ci-dev --locked --all-features --all-targets
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   # https://twitter.com/alcuadrado/status/1571291687837732873
   update:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/rolling.yml
```diff
@@ -6,6 +6,7 @@ env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: FULL
   RUSTFLAGS: "-C link-arg=-fuse-ld=lld"
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dkales: Probably best to just add this to the steps that do need it, otherwise all commands have access to it

fabian1409: are there any that dont need it? as far as i can tell they all do. do mean in case new jobs are added in the future?

dkales: yeah I think only the steps that call cargo need it, there are some apt install ones that do not, also rust-cache etc.

fabian1409: should now only be on the steps that need it (i think cargo fmt and generate-lockfile dont?)

dkales (approved):

PR #395, opened 2025-07-09
feat: also add send_prev and recv_next methods

dkales:


mpc-core/src/protocols/rep3/network.rs
```diff
@@ -8,6 +8,7 @@ use mpc_net::Network;
 use super::id::PartyID;
 
 /// Sends `data` to the next party and receives from the previous party.
+#[inline(always)]
```
fabian1409, line 11: nice addition to add inline! we should add it to the shamir net as well

PR #399, opened 2025-07-16
feat: dont use rayon::join for networking - added std::thread::scope based join functions

fabian1409:


mpc-net/src/lib.rs
```diff
@@ -101,3 +101,169 @@ impl std::fmt::Display for ConnectionStats {
         Ok(())
     }
 }
+
+/// Run 2 network closures
+pub fn join<R0: Send, R1: Send>(
```
dkales, line 107: probably can make these `#[inline(always)]`

mpc-core/src/protocols/rep3_ring/gadgets/sort.rs
```diff
@@ -714,7 +714,7 @@ where
             net.send_and_recv_many(PartyID::ID2, &shuffled, PartyID::ID2)?
         }
         PartyID::ID2 => {
-            let (delta, gamma) = rayon::join(
+            let (delta, gamma) = mpc_net::join(
```
dkales: This one can probably removed...

PR #410, opened 2025-08-08
feat!: plain protogalaxy prover

Cesar199999:


co-noir/protogalaxy/src/protogalaxy_prover_internal.rs
```diff
@@ -0,0 +1,370 @@
...
+        &mut result,
+        Univariate::<C::ScalarField, BATCHED_EXTENDED_LENGTH> {
+            evaluations: [C::ScalarField::ONE; BATCHED_EXTENDED_LENGTH],
+        },
+        alphas.as_slice(),
+    );
+    result
+}
+
+pub(crate) fn evaluate_with_domain_start<F: PrimeField, const SIZE: usize>(
```
florin5f: I think this fits better into univariate.rs

co-noir/protogalaxy/src/protogalaxy_prover_internal.rs
florin5f, line 1: So I think we can just call this on MegaFlavour and make it not generic over L, since we will only use it with Mega (hopefully)

co-noir/ultrahonk/src/plain_prover_flavour.rs
```diff
@@ -52,23 +54,43 @@ pub trait PlainProverFlavour: Default + ProverFlavour {
...
+        running_challenge: &[Univariate<F, SIZE>],
+    );
     fn accumulate_relation_univariates<P: HonkCurve<TranscriptFieldType>>(
         univariate_accumulators: &mut Self::AllRelationAcc<P::ScalarField>,
         extended_edges: &ProverUnivariates<P::ScalarField, Self>,
-        relation_parameters: &RelationParameters<P::ScalarField, Self>,
+        relation_parameters: &RelationParameters<P::ScalarField>,
+        scaling_factor: &P::ScalarField,
+    );
+    fn accumulate_relation_univariates_extended_parameters<
```
florin5f, line 69: so then this function(s) would only be implemented for Mega and not blow up the flavour even more

co-noir/protogalaxy/src/protogalaxy_prover.rs
```diff
@@ -0,0 +1,298 @@
...
+    ) -> HonkProofResult<OinkProverMemory<C, L>> {
+        let oink_prover = Oink::<C, H, L>::new(ZeroKnowledge::No);
+        oink_prover.prove(proving_key, &mut self.transcript)
+    }
+
+    pub(crate) fn run_oink_prover_on_each_incomplete_key(
+        &mut self,
+        proving_keys: &mut Vec<ProvingKey<C, L>>,
+    ) -> HonkProofResult<Vec<OinkProverMemory<C, L>>> {
+        // Asummes accumulator has not been folded
```
florin5f: Assumes

co-noir/protogalaxy/src/protogalaxy_prover.rs
florin5f, line 1: copying from the other file:
So I think we can just call this on MegaFlavour and make it not generic over L, since we will only use it with Mega (hopefully)

co-noir/ultrahonk/src/plain_flavours/plain_mega_flavour.rs
```diff
@@ -128,7 +129,7 @@ fn extend_and_batch_univariates_template<F: PrimeField, const SIZE: usize>(
 impl PlainProverFlavour for MegaFlavour {
     type AllRelationAcc<F: PrimeField> = AllRelationAccMega<F>;
     type AllRelationEvaluations<F: PrimeField> = AllRelationEvaluationsMega<F>;
-    type Alphas<F: PrimeField> = MegaAlphas<F>;
+    type Alpha<F: PrimeField> = F;
```
florin5f: Do we even need this type anymore if it is just a Vec of field elements (I am totally fine with this, just wondering if it is possible to remove it from the flavour entirely?)

co-noir/ultrahonk/src/decider/types.rs
```diff
@@ -237,9 +273,11 @@ impl<C: CurveGroup, L: PlainProverFlavour> VerifierMemory<C, L> {
             beta: verifier_memory.challenges.beta,
             gamma: verifier_memory.challenges.gamma,
             public_input_delta: verifier_memory.public_input_delta,
-            alphas: verifier_memory.challenges.alphas,
-            gate_challenges: Default::default(),
+            // TODO TACEO: How to initialize this?
```
florin5f: in bb it is 0 by default, so I guess it's fine like this

co-noir/ultrahonk/src/plain_flavours/plain_ultra_flavour.rs
```diff
@@ -172,10 +175,19 @@ impl PlainProverFlavour for UltraFlavour {
         )
     }
 
+    fn extend_and_batch_univariates_with_distinct_challenges<F: PrimeField, const SIZE: usize>(
+        _acc: &Self::AllRelationAcc<F>,
+        _result: &mut Univariate<F, SIZE>,
+        _first_term: Univariate<F, SIZE>,
+        _running_challenge: &[Univariate<F, SIZE>],
+    ) {
+        todo!();
```
florin5f: If we do it like I mentioned in the other comment then we wouldnt have these empty impls for Ultra, what do you think about that?

co-noir/ultrahonk/src/decider/relations/auxiliary_relation.rs
```diff
@@ -26,6 +26,7 @@ use num_bigint::BigUint;
  *     5 // RAM consistency sub-relation 3
  * };
  */
+#[cfg(not(feature = "protogalaxy"))]
```
florin5f: Had a talk with the guys and the opinion is that we shouldn't do this as a feature, as this could cause problems in the future when we need both cases. Maybe better do this using const generics, and set it for Ultra/Mega as it was before and in the pg case we call it with the right ones?

co-noir/ultrahonk/src/decider/relations/delta_range_constraint_relation.rs
```diff
@@ -242,4 +274,71 @@ impl<F: PrimeField, L: PlainProverFlavour> Relation<F, L> for DeltaRangeConstrai
 
         univariate_accumulator.r3 += tmp_4;
     }
+
+    fn accumulate_with_extended_parameters<const SIZE: usize>(
+        univariate_accumulator: &mut Self::Acc,
+        input: &ProverUnivariatesSized<F, L, SIZE>,
+        _relation_parameters: &RelationParameters<Univariate<F, SIZE>>,
```
florin5f, line 281: In the cases where we don't use the relation_parameters we can just call the usual accumulate? Then we can reduce the duplicate code at least a bit (I think it's totally fine for the other ones, not worth it to optimize this)

co-noir/protogalaxy/src/protogalaxy_prover_internal.rs
```diff
@@ -0,0 +1,370 @@
...
+    let mut combiner_quotient_evals =
+        vec![C::ScalarField::ZERO; BATCHED_EXTENDED_LENGTH - NUM_KEYS];
+    for point in NUM_KEYS..combiner.evaluations.len() {
+        let idx = point - NUM_KEYS;
+        let point_as_fr = C::ScalarField::from(point as u64);
+        let lagrange_0 = C::ScalarField::ONE - point_as_fr;
+        let vanishing_polynomial = point_as_fr * (point_as_fr - C::ScalarField::ONE);
+        combiner_quotient_evals[idx] = (combiner.evaluations[point]
+            - perturbator_evaluation * lagrange_0)
+            * vanishing_polynomial.inverse().unwrap();
```
florin5f: maybe make these unwraps expects to add a little context

co-noir/protogalaxy/src/protogalaxy_prover_internal.rs
```diff
@@ -0,0 +1,370 @@
...
+        ..
+    } = accumulator_prover_memory;
+
+    let polynomial_size = polys.witness.w_l().len();
+    let mut aggregated_relation_evaluations = vec![C::ScalarField::ZERO; polynomial_size];
+    let mut last_coeff = C::ScalarField::ZERO;
+
+    // TACEO TODO: Barretenberg uses balanced parallelization here with the trace_usage_tracker
+    for i in 0..polynomial_size {
+        let row = polys.get_row(i);
```
florin5f, line 125: is there a way to avoid the get_row (at least in the cloning variant) if we are only using it as a reference?

co-noir/protogalaxy/src/tests.rs
florin5f, line 1: in ultrahonk etc we have a tests directory, maybe do this here as well to be somewhat consistent

co-noir/protogalaxy/src/protogalaxy_prover_internal.rs
Cesar199999: Yes, that makes sense

co-noir/ultrahonk/src/plain_flavours/plain_ultra_flavour.rs
Cesar199999: Definitely agree with you, the only problem I see is that 
```
impl MegaFlavor {
  fn accumulate_relation_univariates_extended_parameters
}
```
yields an error if added to `plain_mega_flavor.rs` (because `MegaFlavor` it is a foreign type). On the other hand, I might run into cyclic deps if if I add it in co-builder.

I'd have to create an additional trait (e.g. `ProtogalaxyMega`) and then `impl ProtogalaxyMega for MegaFlavor`. Sounds good?

co-noir/ultrahonk/src/decider/relations/delta_range_constraint_relation.rs
Cesar199999: Totally, that makes sense, should be fixed now

co-noir/protogalaxy/src/tests.rs
Cesar199999: I thought about this, but it'd force me to make lots of components public. What do you think?

co-noir/ultrahonk/src/plain_flavours/plain_ultra_flavour.rs
florin5f: ah true, that's a problem...
I mean it's just two additional functions, right? maybe we leave it as is for the moment? because having this trait just for these two functions seems also a bit overkill to me

co-noir/ultrahonk/src/plain_flavours/plain_mega_flavour.rs
Cesar199999: Not really, let me try to get rid of it

florin5f (approved):

florin5f (approved):

PR #409, opened 2025-08-06
feat: add plain ECCVM Prover

florin5f:

PR #411, opened 2025-08-08
fix!: grand product computation with inactive regions

Cesar199999:
This PR fixes three inconsistencies with Barretenberg in the grand product computation when dealing with inactive regions

- Virtual index mapping [(line 496)](https://github.com/TaceoLabs/co-snarks/pull/411/files#diff-f0b382c437dac612a98dffcf660341a4e5b08595bcf21bd01b0194dc7bba7c7dR492): Added missing +1 offset when mapping virtual indices to actual z_perm indices. 

- Missing break statement [(line 511)](https://github.com/TaceoLabs/co-snarks/pull/411/files#diff-f0b382c437dac612a98dffcf660341a4e5b08595bcf21bd01b0194dc7bba7c7dR511): Added missing break statement in the inactive region loop to prevent extra iterations.

- Incorrect offset [(line 510)](https://github.com/TaceoLabs/co-snarks/pull/411/files#diff-f0b382c437dac612a98dffcf660341a4e5b08595bcf21bd01b0194dc7bba7c7dR510): Removed +1 offset when accessing z_perm with active range indices. 

Note that prior to this fix, all values within **active ranges** were consistent with those in Barretenberg, which is why existing tests were and remain valid.


Cesar199999:
Linter errors should solve once https://github.com/TaceoLabs/co-snarks/pull/409 is merged

florin5f (approved):

PR #417, opened 2025-08-13
feat!: move witness and input parsing/shareing to new crates for wasm comp

fabian1409:
added new crates that mimic the layout we have in co-circom

noir-types for parsing the noir related types (input, witness, program artifact, ...)
co-noir-types for the new public share types because the acvm types are not supposed to be used for the shared inputs/witnesses and to avoid compiling the acvm crate to wasm

also unified the different public marker enums

moved parsing stuff away from co-solver to new crates


co-noir/co-noir/src/bin/co-noir.rs
```diff
@@ -853,6 +876,26 @@ fn main() -> color_eyre::Result<ExitCode> {
...
+            let constraint_system =
+                co_noir::get_constraint_system_from_artifact(&program_artifact, true);
+            let circuit_size = co_noir::compute_circuit_size::<Bn254G1>(&constraint_system, true)?;
+            let abi = serde_json::to_string(&program_artifact.abi)?;
+            let public_inputs = serde_json::to_string(
+                &program_artifact.bytecode.functions[0]
+                    .public_inputs()
+                    .indices(),
+            )?;
+            tracing::info!("circuit_size = {circuit_size}");
```
0xThemis: nit: should maybe be println if this is the only output for that call. I can theoratically turn off info level and then you dont see anything.
Also for a cli I'd expect that it is not loged but printed.

0xThemis: Also is a json file not better here that you can upload somewhere? if there are 97 public inputs you need to copy them from the terminal which is kinda annoying?

co-noir/co-builder/src/lib.rs
```diff
@@ -51,3 +55,22 @@ pub enum HonkProofError {
...
+    program_artifact: &ProgramArtifact,
+    honk_recusion: bool,
+) -> AcirFormat<ark_bn254::Fr> {
+    let circuit = program_artifact.bytecode.functions[0].to_owned();
+    AcirFormat::circuit_serde_to_acir_format(circuit, honk_recusion)
+}
+
+pub fn constraint_system_from_reader(
+    reader: impl std::io::Read,
+    honk_recusion: bool,
```
0xThemis, line 69: nit: we dont like booleans as parameters. Use an enum Recursion::yes/no

co-noir/co-noir/src/lib.rs
```diff
@@ -39,108 +34,46 @@ pub use co_ultrahonk::{
...
-            PubShared::Shared(f) => {
-                let shares = rep3::share_field_element(f, rng);
-                for (r, share) in res.iter_mut().zip(shares) {
-                    r.push(Rep3AcvmType::from(share));
-                }
-            }
-        }
-    }
-    res
+pub fn split_witness_rep3<P: Pairing>(
```
0xThemis: still think this shoulf be F::Primefield

co-noir/co-noir/src/lib.rs
```diff
@@ -500,16 +418,45 @@ where
...
+    for params in abi.parameters.iter() {
+        let arg_name = &params.name;
+        let typ_field_len = params.typ.field_count();
+        for i in 0..typ_field_len {
+            let should_name = if typ_field_len == 1 {
+                arg_name.to_owned()
+            } else {
+                format!("{arg_name}[{i}]")
+            };
+            let el = witness.get(&should_name).ok_or(eyre::eyre!(
```
0xThemis: nit: use ok_or_else

co-noir/co-ultrahonk/tests/plaindriver.rs
```diff
@@ -29,8 +29,9 @@ fn plaindriver_test<H: TranscriptHasher<TranscriptFieldType>>(
     const CRS_PATH_G1: &str = "../co-builder/src/crs/bn254_g1.dat";
     const CRS_PATH_G2: &str = "../co-builder/src/crs/bn254_g2.dat";
 
-    let constraint_system = Utils::get_constraint_system_from_file(circuit_file, true).unwrap();
-    let witness = Utils::get_witness_from_file(witness_file).unwrap();
+    let constraint_system =
```
0xThemis, line 32: nit:use expect with reason this is ok or use ?

co-noir/noir-types/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    for param in partial_abi.parameters.iter() {
+        let arg_name = &param.name;
+        let typ_field_len = param.typ.field_count();
+
+        // Calculate real witness offset for the public parameter marker
+        loop {
+            let next = orig_params
+                .next()
+                .ok_or(eyre::eyre!("Corrupted Witness: Too few witnesses"))?;
```
0xThemis: or_or_else again
Also if you want to use ok_or there is an ok_or_eyre

co-noir/noir-types/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+        for i in 0..typ_field_len {
+            let name = if typ_field_len == 1 {
+                arg_name.to_owned()
+            } else {
+                format!("{arg_name}[{i}]")
+            };
+
+            let (_, el) = wit_iter
+                .next()
+                .ok_or(eyre::eyre!("Corrupted Witness: Too few witnesses"))?;
```
0xThemis: same as above

co-noir/noir-types/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+            arr.into_iter()
+                .map(json_type_from_value)
+                .collect::<Vec<_>>(),
+        ),
+        serde_json::Value::Object(obj) => JsonTypes::Table(
+            obj.into_iter()
+                .map(|(k, v)| (k, json_type_from_value(v)))
+                .collect(),
+        ),
+        _ => panic!("Unsupported JSON type"),
```
0xThemis: do we want to panic here or return an error? Seems like it should be an error

co-noir/co-noir/src/bin/co-noir.rs
fabian1409: i will stick to printing - you can just pipe it to a file

co-noir/co-ultrahonk/tests/plaindriver.rs
fabian1409: there are dozens of unwraps here bc we are in a test bin

0xThemis (approved):

PR #423, opened 2025-08-24
fix: add runtime curve type check in PlainAcvmSolver::field_shares_to_pointshare

VolodymyrBg:
Fix unsafe downcast in field_shares_to_pointshare by adding runtime curve type validation.

The method was performing unsafe downcasts from ark_grumpkin::Affine to generic C::Affine
without verifying that C is actually the grumpkin curve. This could lead to undefined
behavior if called with a different curve type.

Added runtime TypeId checks to ensure C matches ark_grumpkin::Projective before downcasting,
and renamed variables to eliminate confusing shadowing of 'y'.


florin5f (commented):
Thanks for the PR!
The same check is also done in (I think) 5 other places, would you mind doing it there as well?

VolodymyrBg:
> Thanks for the PR! The same check is also done in (I think) 5 other places, would you mind doing it there as well?

Hm, I found only one place like this. In other functions, we do not cast the grumpkin type to the generic curve C/C::Affine. As it seems to me

florin5f (approved):

PR #433, opened 2025-08-27
feat: add builder type transcript

florin5f:


co-noir/co-builder/src/types/field_ct.rs
```diff
@@ -800,6 +800,21 @@ impl<F: PrimeField> FieldCT<F> {
...
+        T: NoirWitnessExtensionProtocol<P::ScalarField>,
+    >(
+        val: T::AcvmType,
+        driver: &mut T,
+    ) -> eyre::Result<(T::AcvmType, T::AcvmType)> {
+        let is_zero = driver.equal(&val, &F::zero().into())?;
+        let to_invert = driver.cmux(is_zero.to_owned(), F::one().into(), val)?;
+        let inverse = driver.invert(to_invert)?;
+        let selected = driver.cmux(is_zero.to_owned(), F::one().into(), inverse)?;
+        Ok((selected, is_zero))
```
Cesar199999: Maybe I'm wrong but, can't you save a cmux?

```suggestion
        let to_invert = driver.cmux(is_zero.to_owned(), F::one().into(), val)?;
        let inverse = driver.invert(to_invert)?;
        Ok((inverse, is_zero))
```

co-noir/co-builder/src/types/poseidon2.rs
```diff
@@ -414,3 +418,234 @@ impl<F: PrimeField> Default for Poseidon2CT<F, 4, 5> {
...
+        );
+        for input in input.iter() {
+            sponge.absorb(input.clone(), builder, driver)?;
+        }
+
+        let mut output: [FieldCT<P::ScalarField>; OUT_LEN] =
+            array::from_fn(|_| FieldCT::<P::ScalarField>::default());
+        for r in output.iter_mut() {
+            *r = sponge.squeeze(builder, driver)?;
+        }
```
Cesar199999, line 626:
```suggestion
        let mut output: [FieldCT<P::ScalarField>; OUT_LEN] =
            array::from_fn(|_| sponge.squeeze(builder, driver)?);
```

Cesar199999 (dismissed):

co-noir/co-builder/src/types/field_ct.rs
florin5f: you are right, copy paste error without thinking
(also I committed your other suggestion, but then I saw it's not working, so dropping the commit again)

co-noir/co-builder/src/types/poseidon2.rs
florin5f: That's not working for me(?)

Cesar199999: Ah, there's issues with using the sponge in the closure, right? Then forget it :)

PR #461, opened 2025-10-01
fix: incorrect extra multiply in pow_public for field and ring backends

MozirDmitriy:
The square-and-multiply implementation in pow_public performed an extra multiplication after the main loop, returning res * base instead of res. This caused incorrect results (e.g., exponent 1 produced base^3). The issue existed in both field (PrimeField) and ring backends.

Why necessary:
- Ensures exponentiation correctness for shared^public across backends.
- Aligns with the standard square-and-multiply algorithm.
- Prevents silent arithmetic errors in MPC computations and dependent VM operations.

### DIFF
--- a/mpc-core/src/protocols/rep3/arithmetic.rs
+++ b/mpc-core/src/protocols/rep3/arithmetic.rs
@@ -424,7 +424,7 @@ pub fn pow_public<F: PrimeField, N: Network>(
         shared = mul(shared, shared, net, state)?;
         public >>= 1;
     }
-    mul(res, shared, net, state)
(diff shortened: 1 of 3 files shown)


rw0x0:
Hi, thank you for spotting this! In order to merge the PR, can you please fix it in the field backend as well and add a testcase? thank you

MozirDmitriy:
> Hi, thank you for spotting this! In order to merge the PR, can you please fix it in the field backend as well and add a testcase? thank you

Hi, yeah just forgot to make changes to backend, also added test, thanks for review

PR #462, opened 2025-10-02
feat: add u512 IntRing type

florin5f:


mpc-core/src/protocols/rep3/detail.rs
```diff
@@ -269,9 +269,6 @@ fn and_twice<F: PrimeField, N: Network>(
     state: &mut Rep3State,
     bitlen: usize,
 ) -> eyre::Result<(Rep3BigUintShare<F>, Rep3BigUintShare<F>)> {
-    debug_assert!(a.a.bits() <= bitlen as u64);
```
florin5f, line 272: for the rep3_ring::ring_share::rep3_ring_to_field_cast_a2b test had to remove these asserts, how do we want to handle this?

mpc-core/src/protocols/rep3_ring/ring/int_ring.rs
```diff
@@ -280,3 +276,352 @@ impl IntRing2k for u128 {
...
+        U512(Uint::<512, 8>::from(v))
+    }
+}
+impl From<u128> for U512 {
+    fn from(v: u128) -> Self {
+        U512(Uint::<512, 8>::from(v))
+    }
+}
+
+impl From<U512> for u128 {
```
rw0x0: I think we should remove the Into<u128> requirement for the trait (it seems unneeded) und also remove this implementaiton

mpc-core/src/protocols/rep3_ring/ring/int_ring.rs
```diff
@@ -280,3 +276,352 @@ impl IntRing2k for u128 {
...
+impl ShrAssign<usize> for U512 {
+    fn shr_assign(&mut self, rhs: usize) {
+        self.0 >>= rhs;
+    }
+}
+
+impl Neg for U512 {
+    type Output = Self;
+    fn neg(self) -> Self::Output {
+        if self.0.is_zero() {
```
rw0x0: U512(-self.0)

mpc-core/src/protocols/rep3_ring/ring/int_ring.rs
```diff
@@ -280,3 +276,352 @@ impl IntRing2k for u128 {
...
+        let le = v.0.to_le_bytes::<{ U512::BYTES }>();
+        let mut arr = [0u8; 16];
+        arr.copy_from_slice(&le[..16]);
+        u128::from_le_bytes(arr)
+    }
+}
+impl TryFrom<U512> for usize {
+    type Error = &'static str;
+    fn try_from(v: U512) -> Result<Self, Self::Error> {
+        let x: u128 = v.into();
```
rw0x0: This won't work if you remove the impl. you can check directly with if v.0 > usize::MAX and then cut the limbs

mpc-core/src/protocols/rep3_ring/ring/int_ring.rs
```diff
@@ -280,3 +276,352 @@ impl IntRing2k for u128 {
...
+    fn write<W: Write>(&self, mut writer: W) -> std::io::Result<()> {
+        writer.write_all(&self.0.to_le_bytes::<{ Self::BYTES }>())
+    }
+
+    fn bits(&self) -> usize {
+        if self.is_zero() {
+            return 0;
+        }
+        self.0
+            .checked_log2()
```
rw0x0: this can be bit_len() instead of checked_log2(). On the same note, I think the bits() for all other IntRing2k implementations (except Bit) are wrong, there is a +1 missing after ilog2() everywhere, can you fix that as well please?

PR #465, opened 2025-10-08
fix!: incorrect denominator inverses when `num_evals == 1`

Cesar199999:


co-noir/common/src/barycentric.rs
```diff
@@ -49,11 +49,14 @@ impl Barycentric {
 
         let res_size = domain_size * num_evals;
         let mut res = vec![F::zero(); res_size]; // default init to 0 since below does not init all elements
-
-        for k in domain_size..num_evals {
-            for j in 0..domain_size {
-                let inv = lagrange_denominators[j] * (big_domain[k] - big_domain[j]);
-                res[k * domain_size + j] = inv;
+        if num_evals == 1 {
```
rw0x0, line 79: why is this required? The original barretenberg code is (imo) like what we had before:  
for (size_t k = domain_size; k < num_evals; ++k) {
            for (size_t j = 0; j < domain_size; ++j) {
                Fr inv = lagrange_denominators[j];
                inv *= (big_domain[k] - big_domain[j]);
                result[k * domain_size + j] = inv;
            }
        }

rw0x0: as discussed offline, you are right, barretenberg has two versions of this function :)

Cesar199999: c.f. https://github.com/AztecProtocol/barretenberg/blob/4306250af7b46d804168b59b37cec65303acbc63/cpp/src/barretenberg/polynomials/barycentric.hpp#L201

PR #551, opened 2026-06-16
fix(co-acvm): reject nonzero empty assert-zero residuals

### LINKED ISSUE #550: co-acvm accepts fully-known nonzero AssertZero residuals
## Summary

`co-noir/co-acvm/src/solver/assert_zero_solver.rs::solve_assert_zero` treats an empty `linear_combinations` list as success without checking whether the reduced constant residual is zero. A fully-known false `AssertZero` can therefore be accepted by witness generation.

Affected commit checked: `c2765320d9c26a3e8b8f7eb8339c1c9cd9986ad3`.

## Steps to reproduce

Public repro:

https://github.com/Kuhai9801/co-snarks-assertzero-ci/actions/runs/27598641418

The guarded repro checks the audited upstream branch:

```text
if simplified.linear_combinations.is_empty() {
    tracing::trace!("nothing to do for us");
    Ok(())
}
```

then models a fully-known nonzero residual.

## Expected behavior

If an `AssertZero` expression is fully evaluated, the solver should return success only when the residual is zero. A nonzero residual should produce an unsatisfied-constraint error.

## Actual behavior

The empty-linear-combinations branch accepts zero and nonzero residuals the same way.

## Environment

Public GitHub Actions on Ubuntu.

## Notes

The terminal empty-expression case should distinguish a zero residual from a nonzero residual before returning success.

1sgtpepper:
Fixes #550

## Summary

`solve_assert_zero` now checks fully reduced expressions instead of accepting every empty linear combination.

Public zero residuals still succeed. Shared residuals are opened and checked, so valid private constraints that reduce to zero remain accepted, while nonzero public or shared residuals return `UnsatisfiedConstraint`.

## Testing

- `git diff --check`
- Targeted source review of public/shared residual handling

No local compile or test run; verification should run in CI.

### DIFF
--- a/co-noir/co-acvm/src/solver/assert_zero_solver.rs
+++ b/co-noir/co-acvm/src/solver/assert_zero_solver.rs
@@ -145,6 +145,9 @@ where
             if let Some(residual) = T::get_public(&simplified.q_c) {
                 ensure_zero_residual(residual)
             } else if let Some(residual) = T::get_shared(&simplified.q_c) {
+                // Opening is intentional here: valid executions reveal zero, while invalid
+                // executions abort after revealing the failing residual value.
+                // Backends that require private failure values should replace this with a zero-test.
                 let opened = self.driver.open_many(&[residual])?;
                 ensure_single_opened_zero_residual(opened)
             } else if T::is_public_zero(&simplified.q_c) {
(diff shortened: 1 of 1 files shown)


1sgtpepper:
One security-model question before treating this as done.

The new shared residual path opens the residual value to check whether it is zero. That fixes invalid acceptance, but on invalid executions it reveals the failing residual rather than only revealing “constraint failed.”

Is that acceptable for the intended MPC threat model, or should privacy-preserving backends use a zero-test/equality protocol or an explicit capability flag for this behavior?

